PR #3, opened 2018-04-24
Add initial nftnl abstraction

faern:
Here comes the first batch of code for the higher level `nftnl` crate that should safely wrap `nftnl-sys`. This contains the top level objects `Table` and `Chain`. In nftables/netfiler the structure is the following basically:
```
tables
    |-> sets (used to store a collection of items to compare against, such as a list of IPs)
    \-> chains
         \-> rules
              |-> expressions ("tcp dport 80" etc)
              \-> verdict (accept, drop etc)
```
One can read https://wiki.archlinux.org/index.php/Nftables for more details.

Rules, sets and expressions will come an later PRs. But they more or less follow the same design at the moment, so good to understand this small subset and argue about if it has problems or not.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/3)

### DIFF
--- a/nftnl/Cargo.toml
+++ b/nftnl/Cargo.toml
@@ -14,4 +14,6 @@ nftnl-1-0-8 = ["nftnl-sys/nftnl-1-0-8"]
 nftnl-1-0-9 = ["nftnl-sys/nftnl-1-0-9"]
 
 [dependencies]
+error-chain = "0.11"
+libc = "0.2.40"
 nftnl-sys = { path = "../nftnl-sys", version = "0.1" }
--- a/nftnl/Cargo.toml
+++ b/nftnl/Cargo.toml
@@ -8,6 +8,10 @@ repository = "https://github.com/mullvad/nftnl-rs"
 keywords = ["nftables", "nft", "firewall", "iptables", "netfilter"]
 categories = ["network-programming", "os::unix-apis", "api-bindings"]
 
+[features]
+nftnl-1-0-7 = ["nftnl-sys/nftnl-1-0-7"]
+nftnl-1-0-8 = ["nftnl-sys/nftnl-1-0-8"]
+nftnl-1-0-9 = ["nftnl-sys/nftnl-1-0-9"]
 
 [dependencies]
 nftnl-sys = { path = "../nftnl-sys", version = "0.1" }
--- a/nftnl/src/chain.rs
+++ b/nftnl/src/chain.rs
@@ -1,0 +1,103 @@
+use libc;
+use nftnl_sys::{self as sys, c_void};
+
+use std::ffi::CStr;
+
+use Table;
+use {ErrorKind, MsgType, Result};
+
+
+pub type Priority = u32;
+
+/// The netfilter event hooks a chain can register for.
+#[derive(Debug, Copy, Clone)]
+#[repr(u16)]
+pub enum Hook {
+    PreRouting = libc::NF_INET_PRE_ROUTING as u16,
+    In = libc::NF_INET_LOCAL_IN as u16,
+    Forward = libc::NF_INET_FORWARD as u16,
+    Out = libc::NF_INET_LOCAL_OUT as u16,
+    PostRouting = libc::NF_INET_POST_ROUTING as u16,
+}
+
+/// Abstraction of a `nftnl_chain`. Chains reside inside [`Table`]s and they hold `Rule`s.
+///
+/// There are two types of chains, "base chain" and "regular chain". See [`set_hook`] for more
+/// details.
+///
+/// [`Table`]: struct.Table.html
+/// [`set_hook`]: #method.set_hook
+pub struct Chain<'a> {
+    chain: *mut sys::nftnl_chain,
+    table: &'a Table,
+}
+
+impl<'a> Chain<'a> {
+    /// Creates a new chain instance inside the given [`Table`] and with the given name.
+    ///
+    /// [`Table`]: struct.Table.html
+    pub fn new<T: AsRef<CStr>>(name: &T, table: &'a Table) -> Result<Chain<'a>> {
+        unsafe {
+            let chain = sys::nftnl_chain_alloc();
+            if chain.is_null() {
+                bail!(ErrorKind::AllocationError);
+            }
+
+            sys::nftnl_chain_set_str(
+                chain,
+                sys::NFTNL_CHAIN_TABLE as u16,
+                table.get_name().as_ptr(),
+            );
+            sys::nftnl_chain_set_str(chain, sys::NFTNL_CHAIN_NAME as u16, name.as_ref().as_ptr());
+            Ok(Chain { chain, table })
+        }
+    }
+
+    /// Sets the hook and priority for this chain. Without calling this method the chain well
+    /// become a "regular chain" without any hook and will thus not receive any traffic unless
+    /// some rule forward packets to it via goto or jump verdicts.
+    ///
+    /// By calling `set_hook` with a hook the chain that is created will be registered with that
+    /// hook and is thus a "base chain". A "base chain" is an entry point for packets from the
+    /// networking stack.
+    pub fn set_hook(&mut self, hook: Hook, priority: Priority) {
+        unsafe {
+            sys::nftnl_chain_set_u32(self.chain, sys::NFTNL_CHAIN_HOOKNUM as u16, hook as u32);
+            sys::nftnl_chain_set_u32(self.chain, sys::NFTNL_CHAIN_PRIO as u16, priority);
+        }
+    }
+
+    pub fn get_name(&self) -> &CStr {
+        unsafe {
+            let ptr = sys::nftnl_chain_get_str(self.chain, sys::NFTNL_CHAIN_NAME as u16);
+            CStr::from_ptr(ptr)
+        }
+    }
+
+    pub fn get_table(&self) -> &Table {
+        self.table
+    }
+}
+
+unsafe impl<'a> ::NlMsg for Chain<'a> {
+    unsafe fn write(&self, buf: *mut c_void, seq: u32, msg_type: MsgType) {
+        let raw_msg_type = match msg_type {
+            MsgType::Add => libc::NFT_MSG_NEWCHAIN,
+            MsgType::Del => libc::NFT_MSG_DELCHAIN,
+        };
+        let header = sys::nftnl_nlmsg_build_hdr(
+            buf as *mut i8,
+            raw_msg_type as u16,
+            self.table.get_family() as u16,
+            libc::NLM_F_ACK as u16,
+            seq,
+        );
+        sys::nftnl_chain_nlmsg_build_payload(header, self.chain);
+    }
+}
+
+impl<'a> Drop for Chain<'a> {
+    fn drop(&mut self) {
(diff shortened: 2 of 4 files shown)


faern:
Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---

*[nftnl/src/chain.rs, line 41 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/3#-LArM6Yj6e_zQ_rgDax-:-LArM6Yl8VKYT-dJaLnE:blxc88v) ([raw file](https://github.com/mullvad/nftnl-rs/blob/409fb4dfe8e03406db36d33a7cb937ead75a03d8/nftnl/src/chain.rs#L41)):*
> ```Rust
>     pub fn new<T: AsRef<CStr>>(name: &T, table: &'a Table) -> Result<Chain<'a>> {
>         unsafe {
>             let chain = sys::nftnl_chain_alloc();
> ```

The current design is that all abstractions over the C `nftnl` structs allocate them in their constructor and set values on them and then free them on drop. Another design would be to store the values we need to set (in this case the name and optional hook+priority) in this struct and then create the lower `nftnl_chain` inside `NlMsg::write`. Both designs and pros and cons. One pro of this design is that one basically only have to store a pointer to the C struct instead of all the required fields for it.

---

*[nftnl/src/lib.rs, line 20 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/3#-LArMvkf3mUviRml6g8a:-LArMvkf3mUviRml6g8b:b5bwwdi) ([raw file](https://github.com/mullvad/nftnl-rs/blob/409fb4dfe8e03406db36d33a7cb937ead75a03d8/nftnl/src/lib.rs#L20)):*
> ```Rust
> 
> mod chain;
> pub use chain::{Chain, Hook, Priority};
> ```

I re-export the most useful and common types directly to the top of the crate. Because `nftnl::chain::Chain` just feels superfluous when `nftnl::Chain` would be just as easy to understand and less bloated.

---

*[nftnl/src/lib.rs, line 63 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/3#-LArNazhDe_X4S4yjDhI:-LArNazhDe_X4S4yjDhJ:beo2ac2) ([raw file](https://github.com/mullvad/nftnl-rs/blob/409fb4dfe8e03406db36d33a7cb937ead75a03d8/nftnl/src/lib.rs#L63)):*
> ```Rust
> /// 64 KBytes.
> pub fn nft_nlmsg_maxsize() -> u32 {
>     ::std::u16::MAX as u32 + unsafe { libc::sysconf(libc::_SC_PAGESIZE) } as u32
> ```

Just an identical implementation to the one in `libnftnl`. It's not automatically generated for us since it's a macro and bindgen can't translate those.

---

*[nftnl/src/table.rs, line 49 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/3#-LArNnrU2cdYJYfp-ui2:-LArNnrWEFGRFTkXz66c:bsngnlr) ([raw file](https://github.com/mullvad/nftnl-rs/blob/409fb4dfe8e03406db36d33a7cb937ead75a03d8/nftnl/src/table.rs#L49)):*
> ```Rust
>         let raw_msg_type = match msg_type {
>             MsgType::Add => libc::NFT_MSG_NEWTABLE,
>             MsgType::Del => libc::NFT_MSG_DELTABLE,
> ```

When one create a netlink message from a table one can create a "create table" or a "delete table" message. With this kind of implementation one `Table` instance can be used to both generate create and delete messages. Handy when one wants to both delete and then re-create a table or other object in a single batch. I have used it as the following:
```rust
batch.add(&table, nftnl::MsgType::Del)?;
batch.add(&table, nftnl::MsgType::Add)?;
```

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/3)*

pinkisemils:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[nftnl/src/lib.rs, line 24 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/3#-LArMZTxAFlAUmGdZXFO:-LArMZTxAFlAUmGdZXFP:b92j70f) ([raw file](https://github.com/mullvad/nftnl-rs/blob/409fb4dfe8e03406db36d33a7cb937ead75a03d8/nftnl/src/lib.rs#L24)):*
> ```Rust
> 
> #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
> pub enum MsgType {
> ```

`MsgType` can be represented the same way as `ProtoFamily`, can it not? 

---

*[nftnl/src/table.rs, line 29 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/3#-LArVb5P9Zu5_IrgYfqP:-LArVb5P9Zu5_IrgYfqQ:bsfwunu) ([raw file](https://github.com/mullvad/nftnl-rs/blob/409fb4dfe8e03406db36d33a7cb937ead75a03d8/nftnl/src/table.rs#L29)):*
> ```Rust
>             sys::nftnl_table_set_str(table, sys::NFTNL_TABLE_NAME as u16, name.as_ref().as_ptr());
>             sys::nftnl_table_set_u32(table, sys::NFTNL_TABLE_FAMILY as u16, family as u32);
>             Ok(Table { table, family })
> ```

Should `Table` not have a phantom field to attach the lifetime of `name` to the struct, as `name` has to outlive this struct?

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/3)*

faern:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[nftnl/src/lib.rs, line 24 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/3#-LArMZTxAFlAUmGdZXFO:-LArX2Rs3lCtYP5qzE7s:bd91s20) ([raw file](https://github.com/mullvad/nftnl-rs/blob/409fb4dfe8e03406db36d33a7cb937ead75a03d8/nftnl/src/lib.rs#L24)):*
<details><summary><i>Previously, pinkisemils (Emīls Piņķis) wrote…</i></summary><blockquote>

`MsgType` can be represented the same way as `ProtoFamily`, can it not? 
</blockquote></details>

Not really. Because each variant does not correspond to a single raw integer value. It's different for different netfilter objects. Compare the `NlMsg` implementation for `Table` and `Chain`. It's evaluated to `NFT_MSG_NEWTABLE` and `NFT_MSG_NEWCHAIN` respectively.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/3)*

faern:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[nftnl/src/table.rs, line 29 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/3#-LArVb5P9Zu5_IrgYfqP:-LArmz6D0c540XXduenj:b-pqymr9) ([raw file](https://github.com/mullvad/nftnl-rs/blob/409fb4dfe8e03406db36d33a7cb937ead75a03d8/nftnl/src/table.rs#L29)):*
<details><summary><i>Previously, pinkisemils (Emīls Piņķis) wrote…</i></summary><blockquote>

Should `Table` not have a phantom field to attach the lifetime of `name` to the struct, as `name` has to outlive this struct?
</blockquote></details>

The name does not have to outlive the struct. The name passed to this constructor is copied into the C struct with `sys::nftnl_table_set_str(table, sys::NFTNL_TABLE_NAME...);`. And then when accessed we return the `CStr` based on the copy in the C struct.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/3)*

jvff:
Reviewed 3 of 4 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[nftnl/src/lib.rs, line 54 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/3#-LArkKA250to7UnUTaNO:-LArkKA250to7UnUTaNP:b72inzo) ([raw file](https://github.com/mullvad/nftnl-rs/blob/409fb4dfe8e03406db36d33a7cb937ead75a03d8/nftnl/src/lib.rs#L54)):*
> ```Rust
>     /// `nft_nlmsg_maxsize()` bytes. This is not checked by the compiler, which is why this method
>     /// is unsafe.
>     unsafe fn write(&self, buf: *mut c_void, seq: u32, msg_type: MsgType);
> ```

Can't `buf` be a `&mut [u8]` or `&mut [i8]`? You'd need the buffer length and use `slice::from_raw_parts(ptr, len)`. Not sure if it would be enough to remove the `unsafe` qualifier, but the type would be more "Rust friendly".

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/3)*

jvff:
Reviewed 1 of 4 files at r1.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/3)*

faern:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[nftnl/src/lib.rs, line 54 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/3#-LArkKA250to7UnUTaNO:-LArvv33CISULlPBYxcK:b-ecgjm0) ([raw file](https://github.com/mullvad/nftnl-rs/blob/409fb4dfe8e03406db36d33a7cb937ead75a03d8/nftnl/src/lib.rs#L54)):*

Can't `buf` be a `&mut [u8]` or `&mut [i8]`? You'd need the buffer length and use `slice::from_raw_parts(ptr, len)`. Not sure if it would be enough to remove the `unsafe` qualifier, but the type would be more "Rust friendly".
</blockquote></details>

It might look more Rust friendly at this level indeed. But I intended this as a low level FFI trait more or less. If you look at the implementers they have no use of a length. They just pass the void pointer down into `nftnl-sys` functions that don't even accept the length anyway. We just have to trust that `libnftnl` never writes anything longer than `nft_nlmsg_maxsize()`

So I opted for this more low level looking thing since that's how it's going to be used by all implementers anyway.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/3)*

jvff:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[nftnl/src/lib.rs, line 54 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/3#-LArkKA250to7UnUTaNO:-LAz9W6F6Us-O8JRZ88o:b-7fzlwq) ([raw file](https://github.com/mullvad/nftnl-rs/blob/409fb4dfe8e03406db36d33a7cb937ead75a03d8/nftnl/src/lib.rs#L54)):*

It might look more Rust friendly at this level indeed. But I intended this as a low level FFI trait more or less. If you look at the implementers they have no use of a length. They just pass the void pointer down into `nftnl-sys` functions that don't even accept the length anyway. We just have to trust that `libnftnl` never writes anything longer than `nft_nlmsg_maxsize()`

So I opted for this more low level looking thing since that's how it's going to be used by all implementers anyway.
</blockquote></details>

I marked this as resolved, but I would argue that this way allows for some compile-time protection. I'm fine either way, though.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/3)*

PR #4, opened 2018-05-03
Add rule and exprs

faern:
I previously added the `Table` type, which is the top level "container". Under that we already have `Chain` in place. Now comes the things you put in the `Chain`s, `Rule`s. And each rule is built up of `Expression`s

The way nftables/netfilter works is that it's more or less a primitive programming language for loading network packet information into registers and then doing operations on those registers and finally end up in a verdict for the packet.

A rule looking like `tcp dport 987 accept` (any packet having tcp destination port == 987 is accepted) when entered into the `nft` tool is under the hood constructed like this in `libnftnl`:
```rust
// Create a new `Rule` in the chain
let mut rule = Rule::new(&my_chain)?;

// Load the "protocol" field of the IPv4 header into the netfilter register
// (hardcoded to REG_1 for now)
rule.add_expr(nft_expr!(payload ip protocol))?;
// Compare the value in the register (REG_1) with `libc::IPPROTO_TCP as u8`.
// A `Cmp` expression that does not match aborts execution of further expressions
// on this rule and netfilter continues with the next rule.
// If the comparison matches the next expression is executed.
rule.add_expr(nft_expr!(cmp == libc::IPPROTO_TCP as u8))?;

// Load the destination port field from the tcp header into REG_1
rule.add_expr(nft_expr!(payload tcp dport))?;
// Compare the value in REG_1 to `987` in big endian (network endian)
rule.add_expr(nft_expr!(cmp == 987.to_be()))?;

// Add a verdict accept expression. If the execution get this far the packet is accepted.
// The "verdict" expression type is not added in this PR. It comes later.
rule.add_expr(nft_expr!(verdict accept)).unwrap();
```
My highly work in progress attempt at integrating this into `mullvad-daemon` can be viewed on the `linux-firewall-integration` branch. For example, here is how I create the "allow loopback" rules:
https://github.com/mullvad/mullvadvpn-app/blob/linux-firewall-integration/talpid-core/src/firewall/linux/mod.rs#L174

Expressions can be created completely without using the macros. But it becomes very verbose. Which is why I added macros that sort of mimics the semantics one would give to create rules in the `nft` CLI tool. So you understand the difference here is an example:
```rust
nft_expr!(payload tcp dport)

// Is expanded to
nftnl::expr::Payload::Transport(nftnl::expr::TransportHeaderField::Tcp(nftnl::expr::TcpHeaderField::Dport))

// Which, with some sane `use` statements would become:
Payload::Transport(TransportHeaderField::Tcp(TcpHeaderField::Dport))
```

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)

### DIFF
--- a/nftnl-sys/src/lib.rs
+++ b/nftnl-sys/src/lib.rs
@@ -8,7 +8,7 @@
 
 //! Low level FFI bindings to [`libnftnl`]. a userspace library providing a low-level netlink
 //! programming interface (API) to the in-kernel nf_tables subsystem.
-//! 
+//!
 //! See [`nftnl`] for a higher level safe abstraction.
 //!
 //! [`libnftnl`]: https://netfilter.org/projects/libnftnl/
--- a/nftnl/Cargo.toml
+++ b/nftnl/Cargo.toml
@@ -16,4 +16,5 @@ nftnl-1-0-9 = ["nftnl-sys/nftnl-1-0-9"]
 [dependencies]
 error-chain = "0.11"
 libc = "0.2.40"
+log = "0.4"
 nftnl-sys = { path = "../nftnl-sys", version = "0.1" }
--- a/nftnl/src/expr/cmp.rs
+++ b/nftnl/src/expr/cmp.rs
@@ -1,10 +1,10 @@
 use libc;
-use nftnl_sys::{self as sys, c_void};
+use nftnl_sys::{self as sys, c_char, c_void};
 
 use std::borrow::Cow;
 use std::ffi::CString;
-use std::slice;
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+use std::slice;
 
 use super::Expression;
 use {ErrorKind, Result};
@@ -58,7 +58,7 @@ impl<T: ToSlice> Cmp<T> {
 impl<T: ToSlice> Expression for Cmp<T> {
     fn to_expr(&self) -> Result<*mut sys::nftnl_expr> {
         unsafe {
-            let expr = sys::nftnl_expr_alloc(b"cmp\0" as *const _ as *const i8);
+            let expr = sys::nftnl_expr_alloc(b"cmp\0" as *const _ as *const c_char);
             if expr.is_null() {
                 bail!(ErrorKind::AllocationError);
             }
--- a/nftnl/src/expr/cmp.rs
+++ b/nftnl/src/expr/cmp.rs
@@ -1,0 +1,223 @@
+use libc;
+use nftnl_sys::{self as sys, c_void};
+
+use std::borrow::Cow;
+use std::ffi::CString;
+use std::slice;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+
+use super::Expression;
+use {ErrorKind, Result};
+
+/// Comparison operator.
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub enum CmpOp {
+    /// Equals.
+    Eq,
+    /// Not equal.
+    Neq,
+    /// Less than.
+    Lt,
+    /// Less than, or equal.
+    Lte,
+    /// Greater than.
+    Gt,
+    /// Greater than, or equal.
+    Gte,
+}
+
+impl CmpOp {
+    pub fn to_raw(&self) -> u32 {
+        use self::CmpOp::*;
+        match *self {
+            Eq => libc::NFT_CMP_EQ as u32,
+            Neq => libc::NFT_CMP_NEQ as u32,
+            Lt => libc::NFT_CMP_LT as u32,
+            Lte => libc::NFT_CMP_LTE as u32,
+            Gt => libc::NFT_CMP_GT as u32,
+            Gte => libc::NFT_CMP_GTE as u32,
+        }
+    }
+}
+
+
+/// Comparator expression. Allows comparing the content of the netfilter register with any value.
+pub struct Cmp<T: ToSlice> {
+    op: CmpOp,
+    data: T,
+}
+
+impl<T: ToSlice> Cmp<T> {
+    /// Returns a new comparison expression comparing the value loaded in the register with the
+    /// data in `data` using the comparison operator `op`.
+    pub fn new(op: CmpOp, data: T) -> Self {
+        Cmp { op, data }
+    }
+}
+
+impl<T: ToSlice> Expression for Cmp<T> {
+    fn to_expr(&self) -> Result<*mut sys::nftnl_expr> {
+        unsafe {
+            let expr = sys::nftnl_expr_alloc(b"cmp\0" as *const _ as *const i8);
+            if expr.is_null() {
+                bail!(ErrorKind::AllocationError);
+            }
+
+            let data = self.data.to_slice();
+            trace!("Creating a cmp expr comparing with data {:?}", data);
+
+            sys::nftnl_expr_set_u32(
+                expr,
+                sys::NFTNL_EXPR_CMP_SREG as u16,
+                libc::NFT_REG_1 as u32,
+            );
+            sys::nftnl_expr_set_u32(expr, sys::NFTNL_EXPR_CMP_OP as u16, self.op.to_raw());
+            sys::nftnl_expr_set(
+                expr,
+                sys::NFTNL_EXPR_CMP_DATA as u16,
+                data.as_ref() as *const _ as *const c_void,
+                data.len() as u32,
+            );
+
+            Ok(expr)
+        }
+    }
+}
+
+#[macro_export]
+macro_rules! nft_expr_cmp {
+    (==) => {
+        $crate::expr::CmpOp::Eq
+    };
+    (!=) => {
+        $crate::expr::CmpOp::Neq
+    };
+    (<) => {
+        $crate::expr::CmpOp::Lt
+    };
+    (<=) => {
+        $crate::expr::CmpOp::Lte
+    };
+    (>) => {
+        $crate::expr::CmpOp::Gt
+    };
+    (>=) => {
+        $crate::expr::CmpOp::Gte
+    };
+    ($op:tt $data:expr) => {
+        $crate::expr::Cmp::new(nft_expr_cmp!($op), $data)
+    };
+}
+
+
+/// A type that can be converted into a byte buffer.
+pub trait ToSlice {
+    /// Returns the data this type represents.
+    fn to_slice(&self) -> Cow<[u8]>;
+}
+
+impl<'a> ToSlice for [u8; 0] {
+    fn to_slice(&self) -> Cow<[u8]> {
+        Cow::Borrowed(&[])
+    }
+}
+
+impl<'a> ToSlice for &'a [u8] {
+    fn to_slice(&self) -> Cow<[u8]> {
+        Cow::Borrowed(self)
+    }
+}
+
+impl<'a> ToSlice for &'a [u16] {
+    fn to_slice(&self) -> Cow<[u8]> {
+        let ptr = self.as_ptr() as *const u8;
+        let len = self.len() * 2;
+        Cow::Borrowed(unsafe { slice::from_raw_parts(ptr, len) })
+    }
+}
+
+impl ToSlice for IpAddr {
+    fn to_slice(&self) -> Cow<[u8]> {
+        match *self {
+            IpAddr::V4(ref addr) => addr.to_slice(),
+            IpAddr::V6(ref addr) => addr.to_slice(),
+        }
+    }
+}
+
+impl ToSlice for Ipv4Addr {
+    fn to_slice(&self) -> Cow<[u8]> {
+        Cow::Owned(self.octets().to_vec())
+    }
+}
+
+impl ToSlice for Ipv6Addr {
+    fn to_slice(&self) -> Cow<[u8]> {
+        Cow::Owned(self.octets().to_vec())
+    }
+}
+
+impl ToSlice for u8 {
+    fn to_slice(&self) -> Cow<[u8]> {
+        Cow::Owned(vec![*self])
+    }
+}
+
+impl ToSlice for u16 {
+    fn to_slice(&self) -> Cow<[u8]> {
+        let b0 = (*self & 0x00ff) as u8;
+        let b1 = (*self >> 8) as u8;
+        Cow::Owned(vec![b0, b1])
+    }
+}
+
+impl ToSlice for u32 {
+    fn to_slice(&self) -> Cow<[u8]> {
+        let b0 = *self as u8;
+        let b1 = (*self >> 8) as u8;
+        let b2 = (*self >> 16) as u8;
+        let b3 = (*self >> 24) as u8;
+        Cow::Owned(vec![b0, b1, b2, b3])
+    }
+}
+
+impl ToSlice for i32 {
+    fn to_slice(&self) -> Cow<[u8]> {
+        let b0 = *self as u8;
+        let b1 = (*self >> 8) as u8;
+        let b2 = (*self >> 16) as u8;
+        let b3 = (*self >> 24) as u8;
+        Cow::Owned(vec![b0, b1, b2, b3])
+    }
+}
+
+impl<'a> ToSlice for &'a str {
+    fn to_slice(&self) -> Cow<[u8]> {
+        Cow::from(self.as_bytes())
+    }
+}
+
+/// Can be used to compare the value loaded by [`Meta::IifName`] and [`Meta::OifName`]. Please
+/// note that it is faster to check interface index than name.
+///
+/// [`Meta::IifName`]: enum.Meta.html#variant.IifName
+/// [`Meta::OifName`]: enum.Meta.html#variant.OifName
+pub enum InterfaceName {
+    /// Interface name must be exactly the value of the `CString`.
+    Exact(CString),
+    /// Interface name must start with the value of the `CString`.
+    ///
+    /// `InterfaceName::StartingWith("eth")` will look like `eth*` when printed and match against
+    /// `eth0`, `eth1`, ..., `eth99` and so on.
+    StartingWith(CString),
+}
+
+impl<'a> ToSlice for InterfaceName {
+    fn to_slice(&self) -> Cow<[u8]> {
+        let bytes = match *self {
+            InterfaceName::Exact(ref name) => name.as_bytes_with_nul(),
+            InterfaceName::StartingWith(ref name) => name.as_bytes(),
+        };
+        Cow::from(bytes)
+    }
+}
--- a/nftnl/src/expr/cmp.rs
+++ b/nftnl/src/expr/cmp.rs
@@ -86,26 +86,26 @@ impl<T: ToSlice> Expression for Cmp<T> {
 
 #[macro_export]
 macro_rules! nft_expr_cmp {
-    (==) => {
+    (@cmp_op ==) => {
         $crate::expr::CmpOp::Eq
     };
-    (!=) => {
+    (@cmp_op !=) => {
         $crate::expr::CmpOp::Neq
     };
-    (<) => {
+    (@cmp_op <) => {
         $crate::expr::CmpOp::Lt
     };
-    (<=) => {
+    (@cmp_op <=) => {
         $crate::expr::CmpOp::Lte
     };
-    (>) => {
+    (@cmp_op >) => {
         $crate::expr::CmpOp::Gt
     };
-    (>=) => {
+    (@cmp_op >=) => {
         $crate::expr::CmpOp::Gte
     };
     ($op:tt $data:expr) => {
-        $crate::expr::Cmp::new(nft_expr_cmp!($op), $data)
+        $crate::expr::Cmp::new(nft_expr_cmp!(@cmp_op $op), $data)
     };
 }
 
--- a/nftnl/src/expr/counter.rs
+++ b/nftnl/src/expr/counter.rs
@@ -17,10 +17,3 @@ impl Expression for Counter {
         }
     }
 }
-
-#[macro_export]
-macro_rules! nft_expr_counter {
-    () => {
-        $crate::expr::Counter
-    };
-}
--- a/nftnl/src/expr/counter.rs
+++ b/nftnl/src/expr/counter.rs
@@ -1,5 +1,5 @@
 use super::Expression;
-use nftnl_sys as sys;
+use nftnl_sys::{self as sys, c_char};
 use {ErrorKind, Result};
 
 /// A counter expression adds a counter to the rule that is incremented to count number of packets
@@ -9,7 +9,7 @@ pub struct Counter;
 impl Expression for Counter {
     fn to_expr(&self) -> Result<*mut sys::nftnl_expr> {
         unsafe {
-            let expr = sys::nftnl_expr_alloc(b"counter\0" as *const _ as *const i8);
+            let expr = sys::nftnl_expr_alloc(b"counter\0" as *const _ as *const c_char);
             if expr.is_null() {
                 bail!(ErrorKind::AllocationError);
             }
--- a/nftnl/src/expr/counter.rs
+++ b/nftnl/src/expr/counter.rs
@@ -1,0 +1,26 @@
+use super::Expression;
+use nftnl_sys as sys;
+use {ErrorKind, Result};
+
+/// A counter expression adds a counter to the rule that is incremented to count number of packets
+/// and number of bytes for all packets that has matched the rule.
+pub struct Counter;
+
+impl Expression for Counter {
+    fn to_expr(&self) -> Result<*mut sys::nftnl_expr> {
+        unsafe {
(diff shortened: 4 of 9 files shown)


faern:
One thing that is very cumbersome and error prone with this solution on expressions is that it's very easy to create expressions that are invalid or don't make sense. For example:

```rust
rule.add_expr(nft_expr!(payload tcp dport))?;
rule.add_expr(nft_expr!(cmp == 987u16))?;
```
Will compare the big endian port number with the host endian (usually little endian) constant 987. So the comparison would fail on all non-big endian computers. The person creating the statement must know the endianness of the value they loaded in the previous expression.

Another problem is the lengths of the different values:
```rust
rule.add_expr(nft_expr!(payload ip daddr))?;
rule.add_expr(nft_expr!(cmp == 1u8))?;
```
Would load the four byte IP address into the register and then compare it with the one byte value `1`. To do this is not really an error per se, but i will not yield the desired outcome.

To mitigate this we must invent some way that one can stack expressions on each other and they know how they are compatible with each other. Maybe some way with traits to make it so that after an `nft_expr!(payload tcp dport)` one can only attach an expression reading a `u16` from the registers and that is aware of the endianess of it.

I did not spend too much time on this yet since we have to make this work fairly soon. I don't think now is the time to invent the perfect system for this. I just wanted to highlight the potential problems. I think we should make it work in the simplest way possible and integrate it in our VPN client and ship it. Then we can make this library more awesome later if we have time and the motivation.

---

Review status: 0 of 8 files reviewed at latest revision, all discussions resolved.

---

*[a discussion](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LB_lQB43h0u29TlL0JW:-LB_lQB43h0u29TlL0JX:b-futf5) (no related file):*
See `man nft` for more details on the different expressions and how all of this works.

---

*[nftnl/src/expr/cmp.rs, line 72 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LB_kncd2EcUeSrrM3Gy:-LB_kncd2EcUeSrrM3Gz:b-qq41sb) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/cmp.rs#L72)):*
> ```Rust
>                 expr,
>                 sys::NFTNL_EXPR_CMP_SREG as u16,
>                 libc::NFT_REG_1 as u32,
> ```

I have hardcoded all expressions to use only `NFT_REG_1` which is the first register. It's possible to use more registers in netfilter and to load values into multiple registers at once etc. But it seems possible to create all combinations of rules that we need by only using the first register. So I did not spend any time on making this more advanced than it had to be.

---

*[nftnl/src/expr/counter.rs, line 7 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LB_lAj45fb2FDwQWmbk:-LB_lAj45fb2FDwQWmbl:bf8dqza) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/counter.rs#L7)):*
> ```Rust
> /// A counter expression adds a counter to the rule that is incremented to count number of packets
> /// and number of bytes for all packets that has matched the rule.
> pub struct Counter;
> ```

This expression is very useful when debugging. Because it makes it easy to see if rules are hit with packets or not.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-:-LB_j3QW9wFO-NKCrXBd:bfruqex)*

jvff:
Reviewed 8 of 8 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[nftnl/src/rule.rs, line 69 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa1kMBEXmGdl9N7nJ8:-LBa1kMBEXmGdl9N7nJ9:b-dtrlkh) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/rule.rs#L69)):*
> ```Rust
>             MsgType::Del => libc::NFT_MSG_DELRULE,
>         };
>         let nlh = sys::nftnl_nlmsg_build_hdr(
> ```

Minor nitpick: `nlh` seems a bit cryptic to me. Maybe calling it `header` makes it clearer?

---

*[nftnl/src/expr/cmp.rs, line 75 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa6m432bnQL8qqF4NM:-LBa6m432bnQL8qqF4NN:bnrls7g) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/cmp.rs#L75)):*
> ```Rust
>             );
>             sys::nftnl_expr_set_u32(expr, sys::NFTNL_EXPR_CMP_OP as u16, self.op.to_raw());
>             sys::nftnl_expr_set(
> ```

Just to be sure, the contents of `data` are copied to a buffer inside netfilter, right?

---

*[nftnl/src/expr/cmp.rs, line 88 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa78bL14jeOx7gNicv:-LBa78bL14jeOx7gNicw:bemxzk0) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/cmp.rs#L88)):*
> ```Rust
> 
> #[macro_export]
> macro_rules! nft_expr_cmp {
> ```

Wouldn't it be better to split this macro in two? Maybe `nft_expr_cmp` and `nft_expr_cmp_op`?

---

*[nftnl/src/expr/cmp.rs, line 212 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa9DvaBOHF6Rd67-0Z:-LBa9DvaBOHF6Rd67-0_:bgko9lu) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/cmp.rs#L212)):*
> ```Rust
>     /// `InterfaceName::Wildcard("eth")` will look like `eth*` when printed and match against
>     /// `eth0`, `eth1`, ..., `eth99` and so on.
>     Wildcard(CString),
> ```

Another naming nitpick. `Wildcard` sounds too generic, doesn't `StartingWith` fill the purpose better?

---

*[nftnl/src/expr/counter.rs, line 12 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa3PyNDwfEL7IUIvpA:-LBa3PyP3LskEHY5Wntt:b-qm5vyh) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/counter.rs#L12)):*
> ```Rust
>     fn to_expr(&self) -> Result<*mut sys::nftnl_expr> {
>         unsafe {
>             let expr = sys::nftnl_expr_alloc(b"counter\0" as *const _ as *const i8);
> ```

I'm not sure I understood this type cast. Why isn't `b"counter\0" as *const c_char` used?

---

*[nftnl/src/expr/payload.rs, line 73 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBaAWav3-NL_UVzTSvI:-LBaAWav3-NL_UVzTSvJ:b-nes143) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/payload.rs#L73)):*
> ```Rust
> }
> 
> // #[derive(Copy, Clone, Eq, PartialEq)]
> ```

Shouldn't this either be uncommented or removed from the review?

---

*[nftnl/src/expr/payload.rs, line 125 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBaCngd5utph8bNN4tH:-LBaCngd5utph8bNN4tI:b-w7it94) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/payload.rs#L125)):*
> ```Rust
> 
> #[derive(Copy, Clone, Eq, PartialEq)]
> pub enum IpHeaderField {
> ```

Is this specific to IPv4, or is it shared between IPv4 and IPv6?

If it is specific to IPv4, doesn't IPv6 have a protocol field as well?
If it is shared, what happens you use an IPv4 address for IPv6?

---

*[nftnl/src/expr/payload.rs, line 223 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBaDgyj56aLl3yrVrIh:-LBaDgyj56aLl3yrVrIi:bkqnspg) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/payload.rs#L223)):*
> ```Rust
>     }
> 
>     fn len(&self) -> u32 {
> ```

Is it risky to just return `2`? Will there be more variants in the future?

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)*

faern:
Review status: 2 of 9 files reviewed at latest revision, 11 unresolved discussions.

---

*[nftnl/src/rule.rs, line 69 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa1kMBEXmGdl9N7nJ8:-LBeRsPN-Gk7paHXGo39:b9s8drv) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/rule.rs#L69)):*

Minor nitpick: `nlh` seems a bit cryptic to me. Maybe calling it `header` makes it clearer?
</blockquote></details>

Good feedback. I don't like cryptic abbreviations myself. This is left from copy pasting from some example :)

---

*[nftnl/src/expr/cmp.rs, line 88 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa78bL14jeOx7gNicv:-LBeT--i46yjLlyXV5NX:btpuc8y) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/cmp.rs#L88)):*

Wouldn't it be better to split this macro in two? Maybe `nft_expr_cmp` and `nft_expr_cmp_op`?
</blockquote></details>

I was thinking doing the inverse would be good. I wanted to export as few macros as possible from this crate. Optimally I think it should only export `nft_expr`, but that would be too messy so I needed one extra per expression type. Actually, the simpler expressions could probably have their logic baked directly into `nft_expr`.

Macros pollute a single global namespace, so I don't want to export like 10 macros or something from this crate.

---

*[nftnl/src/expr/cmp.rs, line 212 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa9DvaBOHF6Rd67-0Z:-LBeTsSD3wl-II2PxIU_:b-kw6mln) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/cmp.rs#L212)):*

Another naming nitpick. `Wildcard` sounds too generic, doesn't `StartingWith` fill the purpose better?
</blockquote></details>

Yes, that is a better name. Changed.

---

*[nftnl/src/expr/counter.rs, line 12 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa3PyNDwfEL7IUIvpA:-LBeYw5J8pHJmc1yBx7e:bh5w8id) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/counter.rs#L12)):*

I'm not sure I understood this type cast. Why isn't `b"counter\0" as *const c_char` used?
</blockquote></details>

Yeah, I should use the `c_char` alias instead of `i8`. Mostly lazily of me. Good that you spot the not cleaned up parts from when I did early experiments with this :ok_hand: 

---

*[nftnl/src/expr/payload.rs, line 125 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBaCngd5utph8bNN4tH:-LBeZ8yH58QhNMdEDcEz:b-vbc4h7) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/payload.rs#L125)):*

Is this specific to IPv4, or is it shared between IPv4 and IPv6?

If it is specific to IPv4, doesn't IPv6 have a protocol field as well?
If it is shared, what happens you use an IPv4 address for IPv6?
</blockquote></details>

It's IPv4 only. It's just that netfilter/nftables calls the two protocols `ip` and `ip6` respectively so I did stick to that.

There is a separate `Ip6HeaderField` enum below.

Do you think we should stick to netfilter terminology or rename them to `Ipv4HeaderField`/`Ipv6HeaderField`? I don't think I have a strong opinion myself.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)*

pinkisemils:
Reviewed 7 of 8 files at r1, 7 of 7 files at r2.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[nftnl/src/expr/payload.rs, line 125 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBaCngd5utph8bNN4tH:-LBe_mIh1FuAg4Yt4Qvb:b6tdpoa) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/payload.rs#L125)):*

It's IPv4 only. It's just that netfilter/nftables calls the two protocols `ip` and `ip6` respectively so I did stick to that.

There is a separate `Ip6HeaderField` enum below.

Do you think we should stick to netfilter terminology or rename them to `Ipv4HeaderField`/`Ipv6HeaderField`? I don't think I have a strong opinion myself.
</blockquote></details>

I'd prefer`Ipv4HeaderField`/`Ipv6HeaderField`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)*

faern:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[nftnl/src/expr/cmp.rs, line 75 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa6m432bnQL8qqF4NM:-LBehsXo5V0urA7RrnM_:b8tz81j) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/cmp.rs#L75)):*

Just to be sure, the contents of `data` are copied to a buffer inside netfilter, right?
</blockquote></details>

I thought so. But I don't find proof now. Trying to dig down in the library to verify this...

---

*[nftnl/src/expr/payload.rs, line 73 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBaAWav3-NL_UVzTSvI:-LBehfSyD_7bVqg7TE0e:b-i7uc8x) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/payload.rs#L73)):*

Shouldn't this either be uncommented or removed from the review?
</blockquote></details>

Yeah I was not sure. I could just uncomment it. But I have not tested it myself yet. It should be correct, but is untested. I'll uncomment it and test it so we can add it.

---

*[nftnl/src/expr/payload.rs, line 223 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBaDgyj56aLl3yrVrIh:-LBeZnSB3Ng_zNv0qFZ0:b-jv83a6) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/payload.rs#L223)):*

Is it risky to just return `2`? Will there be more variants in the future?
</blockquote></details>

Yes there will be more variants here. This enum represents all the fields of the TCP header, It's just that I did not implement more than the ports yet.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)*

faern:
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---

*[nftnl/src/expr/payload.rs, line 125 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBaCngd5utph8bNN4tH:-LBejtnB1tncFIkXGA5f:bsko7gi) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/payload.rs#L125)):*
<details><summary><i>Previously, pinkisemils (Emīls Piņķis) wrote…</i></summary><blockquote>

I'd prefer`Ipv4HeaderField`/`Ipv6HeaderField`.
</blockquote></details>

Then we need to be consistent in this crate and always follow that naming convention. I'll see if I find any other places where it's used.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)*

faern:
Review status: 8 of 9 files reviewed at latest revision, 11 unresolved discussions.

---

*[nftnl/src/expr/payload.rs, line 125 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBaCngd5utph8bNN4tH:-LBekYSZ0QVG25aQxGpu:b-w42xfg) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/payload.rs#L125)):*

Then we need to be consistent in this crate and always follow that naming convention. I'll see if I find any other places where it's used.
</blockquote></details>

Also had to rename the variants in `NetworkHeaderField` and the identifiers in the macro. But now I think it's consistent.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)*

jvff:
Reviewed 5 of 7 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[nftnl/src/expr/cmp.rs, line 88 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa78bL14jeOx7gNicv:-LBfA6PNAUR0yzIZYyA6:bfjrd57) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/cmp.rs#L88)):*

I was thinking doing the inverse would be good. I wanted to export as few macros as possible from this crate. Optimally I think it should only export `nft_expr`, but that would be too messy so I needed one extra per expression type. Actually, the simpler expressions could probably have their logic baked directly into `nft_expr`.

Macros pollute a single global namespace, so I don't want to export like 10 macros or something from this crate.
</blockquote></details>

In that case, I would suggest to add a marker to "non-public" or "internal" rules. Maybe something like: `(@op ==) => { ... }`, and using it like:

```
($op:tt $data:expr) => {
        $crate::expr::Cmp::new(nft_expr_cmp!(@op $op), $data)
```

Maybe we might want to bikeshed `@op` into something clearer, like `@get_op` or `@into_op`.

I don't think this is actually necessary, just a precaution to avoid:
- user being able to write code using an internal macro rule accidentally and having trouble finding it later;
- leaving open the possibility that the rule might become public in the future (maybe not `==`, but what if `$_:tt` is used?).

Don't mind if you leave it like this though :)

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)*

faern:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[nftnl/src/expr/cmp.rs, line 88 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa78bL14jeOx7gNicv:-LBfIIwaEkO2b5uFUehd:b-unur8v) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/cmp.rs#L88)):*

In that case, I would suggest to add a marker to "non-public" or "internal" rules. Maybe something like: `(@op ==) => { ... }`, and using it like:

```
($op:tt $data:expr) => {
        $crate::expr::Cmp::new(nft_expr_cmp!(@op $op), $data)
```

Maybe we might want to bikeshed `@op` into something clearer, like `@get_op` or `@into_op`.

I don't think this is actually necessary, just a precaution to avoid:
- user being able to write code using an internal macro rule accidentally and having trouble finding it later;
- leaving open the possibility that the rule might become public in the future (maybe not `==`, but what if `$_:tt` is used?).

Don't mind if you leave it like this though :)
</blockquote></details>

I like that idea. Is the `@` prefix something standard you have seen? As you can see I have stuff such as `(ipv6_field saddr) => { $crate::expr::Ipv6HeaderField::Saddr }` in other places, where I have the `ipv6_field` ident to make it internal, but no `@`.

I want to invent as little own style as possible, so if `@` is commonly used I will use it, if not I will skip it.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)*

jvff:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[nftnl/src/expr/cmp.rs, line 88 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa78bL14jeOx7gNicv:-LBg0fCw6yaRT0zysrgC:bk8szt9) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/cmp.rs#L88)):*

I like that idea. Is the `@` prefix something standard you have seen? As you can see I have stuff such as `(ipv6_field saddr) => { $crate::expr::Ipv6HeaderField::Saddr }` in other places, where I have the `ipv6_field` ident to make it internal, but no `@`.

I want to invent as little own style as possible, so if `@` is commonly used I will use it, if not I will skip it.
</blockquote></details>

I've seen it in some other macros, when the macro is used as a "stack-machine", the `@` is used as a marker. Can't remember where I first saw it, but this looks like a good reference: https://danielkeep.github.io/tlborm/book/pat-internal-rules.html

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)*

faern:
Review status: 5 of 9 files reviewed at latest revision, 4 unresolved discussions.

---

*[nftnl/src/expr/cmp.rs, line 88 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa78bL14jeOx7gNicv:-LBtUpge4BRDF3hW7cuH:bsz7u5a) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/cmp.rs#L88)):*

I've seen it in some other macros, when the macro is used as a "stack-machine", the `@` is used as a marker. Can't remember where I first saw it, but this looks like a good reference: https://danielkeep.github.io/tlborm/book/pat-internal-rules.html
</blockquote></details>

Awesome. At the same time that reference is a great argument towards why we should merge the macros. I'll switch to `@`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)*

pinkisemils:
Reviewed 1 of 8 files at r1, 1 of 1 files at r3, 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)*

jvff:
Reviewed 3 of 4 files at r4.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)*

faern:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[nftnl/src/expr/cmp.rs, line 75 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/4#-LBa6m432bnQL8qqF4NM:-LC2huMO4If5PKqXwQ71:b-1rlpif) ([raw file](https://github.com/mullvad/nftnl-rs/blob/97472f22eadd0e5c83be0bfe1760eb0d254403ce/nftnl/src/expr/cmp.rs#L75)):*

I thought so. But I don't find proof now. Trying to dig down in the library to verify this...
</blockquote></details>

I have now traced the executions of setting a few rules in the C `nft` tool. Inside `nftnl_expr_set` it calls to `expr->ops->set(expr, type, data, data_len)` meaning each type of expression define their own set method. I have not checked all expressions, but all the ones I did check did copy the value. So I would be surprised if suddenly a few didn't.

And yes, the setter for the `cmp` expression (which is the one this question is attached to) copies the data with `memcpy`.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)*

jvff:
Reviewed 1 of 4 files at r4.

---



*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/4)*

PR #5, opened 2018-05-15
Add Batch and FinalizedBatch

faern:
Add `Batch` support. A `Batch` is a series of netlink messages that are processed atomically. Thus allowing us to replace an entire ruleset with a new one without any inconsistent state in between.

Batches are natively supported by `libnftnl`, so this is basically just a safe Rust wrapper on top of it.

We force the addition of the begin batch message upon `Batch` creation and we force the addition of the end batch message before converting to a `FinalizedBatch`. Since `FinalizedBatch` is the only type allowing you to get the actual netlink messages out it becomes possible to send a batch without proper begin and end messages.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/5)

### DIFF
--- a/nftnl/src/batch.rs
+++ b/nftnl/src/batch.rs
@@ -1,0 +1,148 @@
+use libc;
+use nftnl_sys::{self as sys, c_void};
+
+use std::ptr;
+use {ErrorKind, MsgType, NlMsg, Result};
+
+/// Check if the kernel supports batched netlink messages to netfilter.
+pub fn batch_is_supported() -> Result<bool> {
+    match unsafe { sys::nftnl_batch_is_supported() } {
+        1 => Ok(true),
+        0 => Ok(false),
+        _ => bail!(ErrorKind::NetlinkError),
+    }
+}
+
+pub struct Batch {
+    batch: *mut sys::nftnl_batch,
+    seq: u32,
+}
+
+impl Batch {
+    /// Creates a new nftnl batch with the [default page size]
+    ///
+    /// [default page size]: fn.default_batch_page_size.html
+    pub fn new() -> Result<Self> {
+        Self::with_page_size(default_batch_page_size())
+    }
+
+    pub fn with_page_size(batch_page_size: u32) -> Result<Self> {
+        let batch = unsafe { sys::nftnl_batch_alloc(batch_page_size, ::nft_nlmsg_maxsize()) };
+        if batch.is_null() {
+            bail!(ErrorKind::AllocationError);
+        }
+        let mut this = Batch { batch, seq: 1 };
+        this.write_begin_msg()?;
+        Ok(this)
+    }
+
+    pub fn add<T: NlMsg>(&mut self, msg: &T, msg_type: MsgType) -> Result<()> {
+        trace!("Writing NlMsg with seq {} to batch", self.seq);
+        unsafe { msg.write(self.current(), self.seq, msg_type) };
+        self.next()
+    }
+
+    pub fn add_iter<T, I>(&mut self, msg_iter: I, msg_type: MsgType) -> Result<()>
+    where
+        T: NlMsg,
+        I: Iterator<Item = T>,
+    {
+        for msg in msg_iter {
+            self.add(&msg, msg_type)?;
+        }
+        Ok(())
+    }
+
+    pub fn finalize(mut self) -> Result<FinalizedBatch> {
+        self.write_end_msg()?;
+        Ok(FinalizedBatch { batch: self })
+    }
+
+    fn current(&self) -> *mut c_void {
+        unsafe { sys::nftnl_batch_buffer(self.batch) }
+    }
+
+    fn next(&mut self) -> Result<()> {
+        if unsafe { sys::nftnl_batch_update(self.batch) } < 0 {
+            bail!(ErrorKind::AllocationError);
+        }
+        self.seq += 1;
+        Ok(())
+    }
+
+    fn write_begin_msg(&mut self) -> Result<()> {
+        unsafe { sys::nftnl_batch_begin(self.current() as *mut i8, self.seq) };
+        self.next()
+    }
+
+    fn write_end_msg(&mut self) -> Result<()> {
+        unsafe { sys::nftnl_batch_end(self.current() as *mut i8, self.seq) };
+        self.next()
+    }
+
+    pub fn as_raw_batch(&self) -> *mut sys::nftnl_batch {
+        self.batch
+    }
+}
+
+impl Drop for Batch {
+    fn drop(&mut self) {
+        unsafe { sys::nftnl_batch_free(self.batch) };
+    }
+}
+
+pub struct FinalizedBatch {
+    batch: Batch,
+}
+
+impl FinalizedBatch {
+    pub fn iter<'a>(&'a self) -> Iter<'a> {
+        let num_pages = unsafe { sys::nftnl_batch_iovec_len(self.batch.as_raw_batch()) as usize };
+        let mut iovecs = vec![
+            libc::iovec {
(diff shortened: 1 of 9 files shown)


jvff:
:lgtm:

Looks awesome! :ok_hand: 

---

Reviewed 2 of 2 files at r1.

---

*[nftnl/src/batch.rs, line 31 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/5#-LC_ZL9BBbIUT9SJXXmj:-LC_ZL9DACY1yYqs6Qey:b-idq9oy) ([raw file](https://github.com/mullvad/nftnl-rs/blob/85c8ff1b947a376adf578b8930f8c2b745043da5/nftnl/src/batch.rs#L31)):*
> ```Rust
>     pub fn with_page_size(batch_page_size: u32) -> Result<Self> {
>         let batch = unsafe { sys::nftnl_batch_alloc(batch_page_size, ::nft_nlmsg_maxsize()) };
>         if batch.is_null() {
> ```

Refactor suggestion (not sure if it's an improvement): `ensure!(!batch.is_null(), ErrorKind::AllocationError);`

---

*[nftnl/src/batch.rs, line 45 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/5#-LC_a2cRCOZQsMTBkuSu:-LC_a2cRCOZQsMTBkuSv:b-aghgl8) ([raw file](https://github.com/mullvad/nftnl-rs/blob/85c8ff1b947a376adf578b8930f8c2b745043da5/nftnl/src/batch.rs#L45)):*
> ```Rust
>     }
> 
>     pub fn add_iter<T, I>(&mut self, msg_iter: I, msg_type: MsgType) -> Result<()>
> ```

Would it make sense to implement `FromIterator<T: NlMsg>` for `Batch`? Or maybe for `FinalizedBatch`? Or maybe even both?

That would allow using `.collect()` on an iterator and get the desired type. Not sure what the general usage case is, so I'm not sure if this suggestion is useful.

---

*[nftnl/src/batch.rs, line 66 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/5#-LC__cGZ9WsMO3zwWj2n:-LC__cGZ9WsMO3zwWj2o:bh3sbsz) ([raw file](https://github.com/mullvad/nftnl-rs/blob/85c8ff1b947a376adf578b8930f8c2b745043da5/nftnl/src/batch.rs#L66)):*
> ```Rust
> 
>     fn next(&mut self) -> Result<()> {
>         if unsafe { sys::nftnl_batch_update(self.batch) } < 0 {
> ```

Same suggestion to use `ensure!` as before, but in both cases it's a purely subjective aesthetic issue.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/5#-:-LC_c2YPC-mdFRhl4_hh:bb18rkz)*

faern:
Thanks :)

I realized a typo in my first comment. It should be "**im**possible to send a batch without proper ..."

---

Review status: 1 of 9 files reviewed at latest revision, 3 unresolved discussions.

---

*[nftnl/src/batch.rs, line 31 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/5#-LC_ZL9BBbIUT9SJXXmj:-LCbvfGg-Em9-tmxRaVO:b-yb8cou) ([raw file](https://github.com/mullvad/nftnl-rs/blob/85c8ff1b947a376adf578b8930f8c2b745043da5/nftnl/src/batch.rs#L31)):*

Refactor suggestion (not sure if it's an improvement): `ensure!(!batch.is_null(), ErrorKind::AllocationError);`
</blockquote></details>

That's a good idea. Changed it in all instances (It's used in many places, not just in the batch)

---

*[nftnl/src/batch.rs, line 45 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/5#-LC_a2cRCOZQsMTBkuSu:-LCbwIxT8I85tcH_IhYc:bv7j560) ([raw file](https://github.com/mullvad/nftnl-rs/blob/85c8ff1b947a376adf578b8930f8c2b745043da5/nftnl/src/batch.rs#L45)):*

Would it make sense to implement `FromIterator<T: NlMsg>` for `Batch`? Or maybe for `FinalizedBatch`? Or maybe even both?

That would allow using `.collect()` on an iterator and get the desired type. Not sure what the general usage case is, so I'm not sure if this suggestion is useful.
</blockquote></details>

One problem is that `FromIterator` is infallible. Adding one, or many, messages to a batch can fail. One could implement it in such a way that overflowing the batch panics and it would be safe to use for reasonably short lists of messages. But on the other hand we could just ignore that and make people use this which is safer.

---

*[nftnl/src/batch.rs, line 66 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/5#-LC__cGZ9WsMO3zwWj2n:-LCbtn5DAYGE8M6QfwN3:b6wzqjh) ([raw file](https://github.com/mullvad/nftnl-rs/blob/85c8ff1b947a376adf578b8930f8c2b745043da5/nftnl/src/batch.rs#L66)):*

Same suggestion to use `ensure!` as before, but in both cases it's a purely subjective aesthetic issue.
</blockquote></details>

I like it in the `is_null` case, because it becomes compact and short. But here I think it's better to keep it as is. The ensure would be long and clunky with such a long condition IMO. If I change this to `ensure!` it actually becomes four lines of code after formatting. One line more than this design :P

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/5#-:-LCbt-RJ63W7yPof63IS:b-1cs3hx)*

jvff:
Reviewed 8 of 8 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[nftnl/src/batch.rs, line 45 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/5#-LC_a2cRCOZQsMTBkuSu:-LCcJFyl6wXBvYM7F4SL:bjqle59) ([raw file](https://github.com/mullvad/nftnl-rs/blob/85c8ff1b947a376adf578b8930f8c2b745043da5/nftnl/src/batch.rs#L45)):*

One problem is that `FromIterator` is infallible. Adding one, or many, messages to a batch can fail. One could implement it in such a way that overflowing the batch panics and it would be safe to use for reasonably short lists of messages. But on the other hand we could just ignore that and make people use this which is safer.
</blockquote></details>

Good point, understood.

---

*[nftnl/src/batch.rs, line 66 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/5#-LC__cGZ9WsMO3zwWj2n:-LCcJQOC1eIhutWXDoEW:bt736ej) ([raw file](https://github.com/mullvad/nftnl-rs/blob/85c8ff1b947a376adf578b8930f8c2b745043da5/nftnl/src/batch.rs#L66)):*

I like it in the `is_null` case, because it becomes compact and short. But here I think it's better to keep it as is. The ensure would be long and clunky with such a long condition IMO. If I change this to `ensure!` it actually becomes four lines of code after formatting. One line more than this design :P
</blockquote></details>

I see. Yeah, in this case it wouldn't be much of a benefit.

---


*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/5)*

pinkisemils:
Reviewed 1 of 2 files at r1, 8 of 8 files at r2.

---



*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/5)*

pinkisemils:
:lgtm:

---

Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/5#-:-LCd2iN20_YeI38rWLRY:bnfp4nl)*

PR #11, opened 2018-06-05
Add build.rs check for LIB{MNL,NFTNL}_LIB_DIR and use to link

faern:
Equivalent to https://github.com/mullvad/mnl-rs/pull/5 but for this library.

Here we have to emit linking information for both `libnftnl` and `libmnl` since `libnftnl` depends on `libmnl`.

For some reason I don't understand it seems to be important in which order they are emitted. If I emit the `libmnl` linking stuff first it won't compile.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/11)

### DIFF
--- a/nftnl-sys/build.rs
+++ b/nftnl-sys/build.rs
@@ -1,5 +1,8 @@
 extern crate pkg_config;
 
+use std::env;
+use std::path::PathBuf;
+
 #[cfg(feature = "nftnl-1-1-0")]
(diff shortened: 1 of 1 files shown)


pinkisemils:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/11)*

PR #14, opened 2018-07-18
Fix warnings emitted by clippy

faern:
Now when clippy is available via rustup lazy me finally got around to run it on a few more libraries. Here I'm cleaning up what it complains about.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/14)

### DIFF
--- a/nftnl/src/batch.rs
+++ b/nftnl/src/batch.rs
@@ -94,7 +94,7 @@ pub struct FinalizedBatch {
 }
 
 impl FinalizedBatch {
-    pub fn iter<'a>(&'a self) -> Iter<'a> {
+    pub fn iter(&self) -> Iter {
         let num_pages = unsafe { sys::nftnl_batch_iovec_len(self.batch.as_raw_batch()) as usize };
         let mut iovecs = vec![
             libc::iovec {
--- a/nftnl/src/expr/cmp.rs
+++ b/nftnl/src/expr/cmp.rs
@@ -27,9 +27,9 @@ pub enum CmpOp {
 }
 
 impl CmpOp {
-    pub fn to_raw(&self) -> u32 {
+    pub fn to_raw(self) -> u32 {
         use self::CmpOp::*;
-        match *self {
+        match self {
             Eq => libc::NFT_CMP_EQ as u32,
             Neq => libc::NFT_CMP_NEQ as u32,
             Lt => libc::NFT_CMP_LT as u32,
--- a/nftnl/src/expr/payload.rs
+++ b/nftnl/src/expr/payload.rs
@@ -19,8 +19,8 @@ pub enum Payload {
 
 
 impl Payload {
-    fn base(&self) -> u32 {
-        match *self {
+    fn base(self) -> u32 {
+        match self {
             // Payload::LinkLayer(_) => libc::NFT_PAYLOAD_LL_HEADER as u32,
             Payload::Network(_) => libc::NFT_PAYLOAD_NETWORK_HEADER as u32,
             Payload::Transport(_) => libc::NFT_PAYLOAD_TRANSPORT_HEADER as u32,
--- a/nftnl/src/lib.rs
+++ b/nftnl/src/lib.rs
@@ -75,5 +75,5 @@ pub unsafe trait NlMsg {
 /// length (nla_len) is 16 bits, the largest message is a bit larger than
 /// 64 KBytes.
 pub fn nft_nlmsg_maxsize() -> u32 {
-    ::std::u16::MAX as u32 + unsafe { libc::sysconf(libc::_SC_PAGESIZE) } as u32
+    u32::from(::std::u16::MAX) + unsafe { libc::sysconf(libc::_SC_PAGESIZE) } as u32
 }
--- a/nftnl/src/rule.rs
+++ b/nftnl/src/rule.rs
@@ -48,7 +48,7 @@ impl<'a> Rule<'a> {
         }
     }
 
-    pub fn add_expr<E: Expression>(&mut self, expr: E) -> Result<()> {
+    pub fn add_expr(&mut self, expr: &impl Expression) -> Result<()> {
         unsafe { sys::nftnl_rule_add_expr(self.rule, expr.to_expr()?) }
         Ok(())
     }


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/14)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @jvff and @pinkisemils)

___
*[nftnl/src/lib.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/14#-LHgBnS0Cl931OaczcMX:-LHgBnS11WROn43txfNc:b-yb94w5) ([raw file](https://github.com/mullvad/nftnl-rs/blob/2ed837ac50b8a4fb94988d5b62527a347a3307ef/nftnl/src/lib.rs#L78)):*
> ```Rust
> /// 64 KBytes.
> pub fn nft_nlmsg_maxsize() -> u32 {
>     ::std::u16::MAX as u32 + unsafe { libc::sysconf(libc::_SC_PAGESIZE) } as u32
> ```

`casting u16 to u32 may become silently lossy if types change`
`help: try: `u32::from(::std::u16::MAX)`
___
*[nftnl/src/rule.rs, line 51 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/14#-LHgCIQdAhOfcQQsgwCh:-LHgCIQdAhOfcQQsgwCi:b-micoqp) ([raw file](https://github.com/mullvad/nftnl-rs/blob/2ed837ac50b8a4fb94988d5b62527a347a3307ef/nftnl/src/rule.rs#L51)):*
> ```Rust
>     }
> 
>     pub fn add_expr<E: Expression>(&mut self, expr: E) -> Result<()> {
> ```

`this argument is passed by value, but not consumed in the function body`
`help: consider taking a reference instead: &E`

But I then decided to also move on to use `impl Trait` since the bound on `E` was simply just a trait bound.
___
*[nftnl/src/expr/cmp.rs, line 30 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/14#-LHgCf4Q1idBAbKhxJIZ:-LHgCf4Q1idBAbKhxJI_:b3ht6gd) ([raw file](https://github.com/mullvad/nftnl-rs/blob/2ed837ac50b8a4fb94988d5b62527a347a3307ef/nftnl/src/expr/cmp.rs#L30)):*
> ```Rust
> 
> impl CmpOp {
>     pub fn to_raw(&self) -> u32 {
> ```

`this argument is passed by reference, but would be more efficient if passed by value`
`note: #[warn(trivially_copy_pass_by_ref)] on by default`
So it doesn't explicitly say so. But from the name of the hint having "trivially copy" in it my guess is that the recommendation is to never take copy types (implementing `Copy`) by reference, rather just take by value, which will not steal ownership, just copy.

jvff (commented):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/14)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @pinkisemils, @jvff, and @faern)

___
*[nftnl/src/lib.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/14#-LHgBnS0Cl931OaczcMX:-LHhFaZp1tBeXW2tzVAk:bplutrf) ([raw file](https://github.com/mullvad/nftnl-rs/blob/2ed837ac50b8a4fb94988d5b62527a347a3307ef/nftnl/src/lib.rs#L78)):*

`casting u16 to u32 may become silently lossy if types change`
`help: try: `u32::from(::std::u16::MAX)`
</blockquote></details>

Can the `::std::` be omitted?
___
*[nftnl/src/rule.rs, line 51 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/14#-LHgCIQdAhOfcQQsgwCh:-LHhFnqI7LzYA-hmRdFQ:b9cz71o) ([raw file](https://github.com/mullvad/nftnl-rs/blob/2ed837ac50b8a4fb94988d5b62527a347a3307ef/nftnl/src/rule.rs#L51)):*

`this argument is passed by value, but not consumed in the function body`
`help: consider taking a reference instead: &E`

But I then decided to also move on to use `impl Trait` since the bound on `E` was simply just a trait bound.
</blockquote></details>

Doesn't this also allocate the virtual call table? Just to be sure on this decision, we're trading extra generated code with an indirect call.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/14)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @pinkisemils, @jvff, and @faern)

___
*[nftnl/src/lib.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/14#-LHgBnS0Cl931OaczcMX:-LHlOItz9dQWmKMd_Gbo:b-4t533s) ([raw file](https://github.com/mullvad/nftnl-rs/blob/2ed837ac50b8a4fb94988d5b62527a347a3307ef/nftnl/src/lib.rs#L78)):*

Can the `::std::` be omitted?
</blockquote></details>

Not without specifically importing it like, `use std::u16` first. And it's only used once, plus the path is not that terrible so I felt it was easier to just have the full path here directly?
___
*[nftnl/src/rule.rs, line 51 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/14#-LHgCIQdAhOfcQQsgwCh:-LHlN8-18rhG5-UtEGRO:bkmkg3d) ([raw file](https://github.com/mullvad/nftnl-rs/blob/2ed837ac50b8a4fb94988d5b62527a347a3307ef/nftnl/src/rule.rs#L51)):*

Doesn't this also allocate the virtual call table? Just to be sure on this decision, we're trading extra generated code with an indirect call.
</blockquote></details>

`impl Trait` is static dispatch/monomorphized. There are no trait objects or dynamic dispatch involved with the `expr` here. `dyn Trait` would be the dynamically dispatched version of that.

`expr: impl MyTrait` is equivalent to having a generic like `<E: MyTrait>(expr: E)`. It's just that argument position `impl Trait`s allow you to get rid of the generic if the only purpose of the generic is to put a trait bound on the type.

pinkisemils (commented):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/14)* status: all files reviewed, 3 unresolved discussions (waiting on @jvff and @faern)

jvff (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/14)* status: all files reviewed, 3 unresolved discussions (waiting on @jvff and @faern)

___
*[nftnl/src/lib.rs, line 78 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/14#-LHgBnS0Cl931OaczcMX:-LHmPuETD8QuVZav4gjM:bmjvkgx) ([raw file](https://github.com/mullvad/nftnl-rs/blob/2ed837ac50b8a4fb94988d5b62527a347a3307ef/nftnl/src/lib.rs#L78)):*

Not without specifically importing it like, `use std::u16` first. And it's only used once, plus the path is not that terrible so I felt it was easier to just have the full path here directly?
</blockquote></details>

Is `u32::from` using `u32` as a primitive and `u16::MAX` using `u16` as a module?

That's confusing :( Now that I understand better, I think it's better to actually not omit it, so it becomes as explicit as possible.
 
Hopefully one day Rust has a `Limited` trait or something with `MIN` and `MAX` associated constants and the module gets merged into the primitive.
___
*[nftnl/src/rule.rs, line 51 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/14#-LHgCIQdAhOfcQQsgwCh:-LHmQnU95_Q2Tswd7BM1:b-m4jkii) ([raw file](https://github.com/mullvad/nftnl-rs/blob/2ed837ac50b8a4fb94988d5b62527a347a3307ef/nftnl/src/rule.rs#L51)):*

`impl Trait` is static dispatch/monomorphized. There are no trait objects or dynamic dispatch involved with the `expr` here. `dyn Trait` would be the dynamically dispatched version of that.

`expr: impl MyTrait` is equivalent to having a generic like `<E: MyTrait>(expr: E)`. It's just that argument position `impl Trait`s allow you to get rid of the generic if the only purpose of the generic is to put a trait bound on the type.
</blockquote></details>

Ah, understood! `impl Trait` is very handy here, cool!

PR #21, opened 2019-04-03
Remove error chain

faern:
Removes the dependency on `error-chain` and replace it with `err-derive` errors.

One odd thing I found was that the `BatchIsFull` error variant was never used. OK, so we forgot to remove it when we stopped using it?... I can't see in the git history that we ever used it. Adding messages to batches has always just returned the allocation error, since adding stuff to batches only fails if `libnftnl` fails to allocate a new page for it.

With `BatchIsFull` removed and the netlink error separated, this brings us down to a single error that can happen. The Rust standard library just panics on allocation errors at the moment. Because there is not very much you can do anyway, so you probably want your thread to exit? So... Should we remove our `AllocationError` and change the `try_alloc!` macro to panic instead? Then usage of the library becomes infallible and the API a lot simpler.

I also did some small maintenance tasks. Updating the formatting to our latest rustfmt.toml version and reformatted. And upgraded a dev dependency.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/21)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 - Upgrade crates to Rust 2018 edition.
-- Remove the `error-chain` dependency and introduce new error types.
+- Remove the `error-chain` dependency. Now panics on allocation error.
 
 
 ## [0.1.0] - 2018-09-10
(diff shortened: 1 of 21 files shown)


pinkisemils (approved):
> Reviewed 21 of 21 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/21)* status: :shipit: complete! all files reviewed, all discussions resolved

pinkisemils (approved):
> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/21)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #23, opened 2019-05-07
Add support for reading ICMPv6 header fields

faern:
For router solicitation/advertisement we need to check the ICMPv6 type and code header fields in our firewall rules. This adds support for ICMPv6 transport header fields.

I also create a rule using this new feature in an example. To prove to myself that it works.

ICMPv6 header spec can be found here: https://en.wikipedia.org/wiki/Internet_Control_Message_Protocol_for_IPv6#Message_types_and_formats

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/23)

### DIFF
--- a/nftnl/examples/add-rules.rs
+++ b/nftnl/examples/add-rules.rs
@@ -146,6 +146,30 @@ fn main() -> Result<(), Error> {
     batch.add(&block_out_to_private_net_rule, nftnl::MsgType::Add);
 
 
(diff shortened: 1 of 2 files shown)


pinkisemils (approved):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/23)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #24, opened 2020-02-25
Add support for cgroups, ct and packet marks, and setting packet metainformation

dlon:
Changes:
* Add expressions `&nft_expr!(meta cgroup)` and `&nft_expr!(meta mark)`
* Set immediate data: `&nft_expr!(immediate data 1234)`
* Set packet metadata, e.g. `&nft_expr!(meta mark set)` (preceded by `immediate ...`)
* Support ct/connection marks: `&nft_expr!(ct mark)` and `&nft_expr!(ct mark set)`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/24)

### DIFF
--- a/nftnl/src/expr/ct.rs
+++ b/nftnl/src/expr/ct.rs
@@ -14,12 +14,14 @@ bitflags::bitflags! {
 
 pub enum Conntrack {
     State,
+    Mark { set: bool },
 }
 
 impl Conntrack {
     fn raw_key(&self) -> u32 {
         match *self {
             Conntrack::State => libc::NFT_CT_STATE as u32,
+            Conntrack::Mark { .. } => libc::NFT_CT_MARK as u32,
         }
     }
 }
@@ -29,7 +31,11 @@ impl Expression for Conntrack {
         unsafe {
             let expr = try_alloc!(sys::nftnl_expr_alloc(b"ct\0" as *const _ as *const c_char));
 
-            sys::nftnl_expr_set_u32(expr, sys::NFTNL_EXPR_CT_DREG as u16, libc::NFT_REG_1 as u32);
+            if let Conntrack::Mark { set: true } = self {
+                sys::nftnl_expr_set_u32(expr, sys::NFTNL_EXPR_CT_SREG as u16, libc::NFT_REG_1 as u32);
+            } else {
+                sys::nftnl_expr_set_u32(expr, sys::NFTNL_EXPR_CT_DREG as u16, libc::NFT_REG_1 as u32);
+            }
             sys::nftnl_expr_set_u32(expr, sys::NFTNL_EXPR_CT_KEY as u16, self.raw_key());
 
             expr
@@ -41,5 +47,11 @@ impl Expression for Conntrack {
 macro_rules! nft_expr_ct {
     (state) => {
         $crate::expr::Conntrack::State
+    };
+    (mark set) => {
+        $crate::expr::Conntrack::Mark { set: true }
+    };
+    (mark) => {
+        $crate::expr::Conntrack::Mark { set: false }
     };
 }
--- a/nftnl/src/expr/immediate.rs
+++ b/nftnl/src/expr/immediate.rs
@@ -1,7 +1,46 @@
 use super::Expression;
 use libc;
-use nftnl_sys::{self as sys, libc::c_char};
-use std::ffi::{CStr, CString};
+use nftnl_sys::{self as sys, libc::{c_char, c_void}};
+use std::{ffi::{CStr, CString}, mem::size_of_val};
+
+/// An immediate expression. Used to set immediate data.
+/// Verdicts are handled separately by [Verdict].
+#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+pub struct Immediate<T> {
+    pub data: T,
+}
+
+impl<T> Expression for Immediate<T> {
+    fn to_expr(&self) -> *mut sys::nftnl_expr {
+        unsafe {
+            let expr = try_alloc!(sys::nftnl_expr_alloc(
+                b"immediate\0" as *const _ as *const c_char
+            ));
+
+            sys::nftnl_expr_set_u32(
+                expr,
+                sys::NFTNL_EXPR_IMM_DREG as u16,
+                libc::NFT_REG_1 as u32,
+            );
+
+            sys::nftnl_expr_set(
+                expr,
+                sys::NFTNL_EXPR_IMM_DATA as u16,
+                &self.data as *const _ as *const c_void,
+                size_of_val(&self.data) as u32,
+            );
+
+            expr
+        }
+    }
+}
+
+#[macro_export]
+macro_rules! nft_expr_immediate {
+    (data $value:expr) => {
+        $crate::expr::Immediate { data: $value }
+    };
+}
 
 /// A verdict expression. In the background actually an "Immediate" expression in nftnl terms,
 /// but here it's simplified to only represent a verdict.
--- a/nftnl/src/expr/meta.rs
+++ b/nftnl/src/expr/meta.rs
@@ -22,6 +22,8 @@ pub enum Meta {
     NfProto,
     /// Layer 4 protocol number.
     L4Proto,
+    /// Socket control group (skb->sk->sk_classid).
+    Cgroup,
     /// A 32bit pseudo-random number
     PRandom,
 }
@@ -40,6 +42,7 @@ impl Meta {
             OifType => libc::NFT_META_OIFTYPE as u32,
             NfProto => libc::NFT_META_NFPROTO as u32,
             L4Proto => libc::NFT_META_L4PROTO as u32,
+            Cgroup => libc::NFT_META_CGROUP as u32,
             PRandom => libc::NFT_META_PRANDOM as u32,
         }
     }
@@ -91,6 +94,9 @@ macro_rules! nft_expr_meta {
     };
     (l4proto) => {
         $crate::expr::Meta::L4Proto
+    };
+    (cgroup) => {
+        $crate::expr::Meta::Cgroup
     };
     (random) => {
         $crate::expr::Meta::PRandom
--- a/nftnl/src/expr/meta.rs
+++ b/nftnl/src/expr/meta.rs
@@ -6,6 +6,8 @@ use nftnl_sys::{self as sys, libc::c_char};
 pub enum Meta {
     /// Packet ethertype protocol (skb->protocol), invalid in OUTPUT.
     Protocol,
+    /// Packet mark.
+    Mark { set: bool },
     /// Packet input interface index (dev->ifindex).
     Iif,
     /// Packet output interface index (dev->ifindex).
@@ -31,9 +33,10 @@ pub enum Meta {
 impl Meta {
     /// Returns the corresponding `NFT_*` constant for this meta expression.
     pub fn to_raw_key(&self) -> u32 {
-        use self::Meta::*;
+        use Meta::*;
         match *self {
             Protocol => libc::NFT_META_PROTOCOL as u32,
+            Mark { .. } => libc::NFT_META_MARK as u32,
             Iif => libc::NFT_META_IIF as u32,
             Oif => libc::NFT_META_OIF as u32,
             IifName => libc::NFT_META_IIFNAME as u32,
@@ -55,11 +58,19 @@ impl Expression for Meta {
                 b"meta\0" as *const _ as *const c_char
             ));
 
-            sys::nftnl_expr_set_u32(
-                expr,
-                sys::NFTNL_EXPR_META_DREG as u16,
-                libc::NFT_REG_1 as u32,
-            );
+            if let Meta::Mark { set: true } = self {
+                sys::nftnl_expr_set_u32(
+                    expr,
+                    sys::NFTNL_EXPR_META_SREG as u16,
+                    libc::NFT_REG_1 as u32,
+                );
+            } else {
+                sys::nftnl_expr_set_u32(
+                    expr,
+                    sys::NFTNL_EXPR_META_DREG as u16,
+                    libc::NFT_REG_1 as u32,
+                );
+            }
             sys::nftnl_expr_set_u32(expr, sys::NFTNL_EXPR_META_KEY as u16, self.to_raw_key());
             expr
         }
@@ -70,6 +81,12 @@ impl Expression for Meta {
 macro_rules! nft_expr_meta {
     (proto) => {
         $crate::expr::Meta::Protocol
+    };
+    (mark set) => {
+        $crate::expr::Meta::Mark { set: true }
+    };
+    (mark) => {
+        $crate::expr::Meta::Mark { set: false }
     };
     (iif) => {
         $crate::expr::Meta::Iif
--- a/nftnl/src/expr/mod.rs
+++ b/nftnl/src/expr/mod.rs
@@ -59,6 +59,9 @@ macro_rules! nft_expr {
     (lookup $set:expr) => {
         nft_expr_lookup!($set)
     };
+    (meta $expr:ident set) => {
+        nft_expr_meta!($expr set)
+    };
     (meta $expr:ident) => {
         nft_expr_meta!($expr)
     };
--- a/nftnl/src/expr/mod.rs
+++ b/nftnl/src/expr/mod.rs
@@ -68,4 +68,7 @@ macro_rules! nft_expr {
     (payload $proto:ident $field:ident) => {
         nft_expr_payload!($proto $field)
     };
+    (immediate $expr:ident $value:expr) => {
+        nft_expr_immediate!($expr $value)
+    };
 }
--- a/nftnl/src/expr/mod.rs
+++ b/nftnl/src/expr/mod.rs
@@ -22,7 +22,7 @@ mod counter;
 pub use self::counter::*;
 
 pub mod ct;
-pub use self::ct::Conntrack;
+pub use self::ct::*;
 
 mod immediate;
 pub use self::immediate::*;
@@ -46,6 +46,9 @@ macro_rules! nft_expr {
     };
     (counter) => {
         $crate::expr::Counter
+    };
+    (ct $key:ident set) => {
+        nft_expr_ct!($key set)
     };
     (ct $key:ident) => {
         nft_expr_ct!($key)


faern (changes requested):
> Reviewed 1 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/24)* status: 1 of 5 files reviewed, 5 unresolved discussions (waiting on @dlon and @faern)

___
*[nftnl/src/expr/ct.rs, line 16 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-fd4rBWkISRbJy9HB:-M1-fd4rBWkISRbJy9HC:b-vntjpu) ([raw file](https://github.com/mullvad/nftnl-rs/blob/744a4282d3fd6d3df7ba4351e6047c0e5598bb04/nftnl/src/expr/ct.rs#L16)):*
> ```Rust
> 
> pub enum Conntrack {
>     Mark,
> ```

Same feedback as on the `Meta` type. Can't we make the enum carry the `set_value` only in the variants where it makes sense? That is if only setting a value makes sense in a subset of the variants. If it's always valid to `set_value: true` my argument does not hold, but if so I think we should work on the structure here to make the wrapper `ConntrackStruct` the top level type and better named and have the variant be a sub-type.
___
*[nftnl/src/expr/immediate.rs, line 6 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-h73p5eH5Bz4Crxpt:-M1-h73p5eH5Bz4Crxpu:b-o9q2zp) ([raw file](https://github.com/mullvad/nftnl-rs/blob/744a4282d3fd6d3df7ba4351e6047c0e5598bb04/nftnl/src/expr/immediate.rs#L6)):*
> ```Rust
> use std::mem::size_of_val;
> 
> /// An immediate expression.
> ```

Since on the `libnftnl` level a verdict is an immediate expression users might come to this module/type to find out how to do verdicts. If we keep them as separate types I think the documentation for the `Immediate` should say this and point to the docs for the `Verdict` type.
___
*[nftnl/src/expr/meta.rs, line 33 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-fCbX6boABwNGroR1:-M1-fCbX6boABwNGroR2:bwpiwnm) ([raw file](https://github.com/mullvad/nftnl-rs/blob/744a4282d3fd6d3df7ba4351e6047c0e5598bb04/nftnl/src/expr/meta.rs#L33)):*
> ```Rust
> }
> 
> pub struct MetaStruct {
> ```

I'm not a fan of naming a struct `Struct`. The only time I think that makes a lot of sense is if you are writing a compiler and your struct actually represent a struct in whatever you are compiling :)
___
*[nftnl/src/expr/meta.rs, line 95 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-em200fgmbaErxIVc:-M1-em212TzfpwCe3c2k:bo41675) ([raw file](https://github.com/mullvad/nftnl-rs/blob/744a4282d3fd6d3df7ba4351e6047c0e5598bb04/nftnl/src/expr/meta.rs#L95)):*
> ```Rust
>     };
>     (@mark $(set)?) => {
>         $crate::expr::Meta::Mark
> ```

Is it *only* mark that can/should be used with `set_value: true`? Then why not have:
```rust
enum Meta {
    ....
    Mark(bool),
    ....
}
```
Instead of creating a new struct that wraps everything.

A goal is that this library should be possible to use cleanly without using the macros. The macros are helpers, but it should be fully possible (and good looking) to use the types directly.
___
*[nftnl/src/expr/verdict.rs, line 9 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-hQ0X1CYwm5bS1eO2:-M1-hQ0X1CYwm5bS1eO3:b-z1056z) ([raw file](https://github.com/mullvad/nftnl-rs/blob/744a4282d3fd6d3df7ba4351e6047c0e5598bb04/nftnl/src/expr/verdict.rs#L9)):*
> ```Rust
> /// but here it's simplified to only represent a verdict.
> #[derive(Debug, Clone, Eq, PartialEq, Hash)]
> pub enum Verdict {
> ```

I'm thinking we should maybe keep this type in `immediate.rs`. So we still have a 1:1 mapping between `libnftnl` expressions and Rust modules. So one can look in a single place to see how we implement the `immediate` expression.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/24)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @faern)

___
*[nftnl/src/expr/ct.rs, line 16 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-fd4rBWkISRbJy9HB:-M1-u90T--3fGSkRDQ6M:bzgkvsh) ([raw file](https://github.com/mullvad/nftnl-rs/blob/744a4282d3fd6d3df7ba4351e6047c0e5598bb04/nftnl/src/expr/ct.rs#L16)):*

Same feedback as on the `Meta` type. Can't we make the enum carry the `set_value` only in the variants where it makes sense? That is if only setting a value makes sense in a subset of the variants. If it's always valid to `set_value: true` my argument does not hold, but if so I think we should work on the structure here to make the wrapper `ConntrackStruct` the top level type and better named and have the variant be a sub-type.
</blockquote></details>

Done. Only `ct mark` can be set, I believe.
___
*[nftnl/src/expr/immediate.rs, line 6 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-h73p5eH5Bz4Crxpt:-M1-uUAUEeXw25eNnTbA:b-896fix) ([raw file](https://github.com/mullvad/nftnl-rs/blob/744a4282d3fd6d3df7ba4351e6047c0e5598bb04/nftnl/src/expr/immediate.rs#L6)):*

Since on the `libnftnl` level a verdict is an immediate expression users might come to this module/type to find out how to do verdicts. If we keep them as separate types I think the documentation for the `Immediate` should say this and point to the docs for the `Verdict` type.
</blockquote></details>

Done.
___
*[nftnl/src/expr/immediate.rs, line 46 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-wNN-8mspSF0PQbs3:-M1-wNN-8mspSF0PQbs4:bj4uxvg) ([raw file](https://github.com/mullvad/nftnl-rs/blob/a94eeb6185b44457ee2a06f665fbc9e274ccd700/nftnl/src/expr/immediate.rs#L46)):*
> ```Rust
> macro_rules! nft_expr_immediate {
>     (data $value:expr) => {
>         $crate::expr::Immediate::Data($value)
> ```

The syntax here is `immediate data <value>`. I'm thinking it should be simply `immediate <value>`. That would be more consistent with the netlink instructions:

```
[ immediate reg 1 <value> ]
[ meta set mark with reg 1 ]
```
___
*[nftnl/src/expr/meta.rs, line 33 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-fCbX6boABwNGroR1:-M1-vQY85IKmp0o094tT:bss4a2b) ([raw file](https://github.com/mullvad/nftnl-rs/blob/744a4282d3fd6d3df7ba4351e6047c0e5598bb04/nftnl/src/expr/meta.rs#L33)):*

I'm not a fan of naming a struct `Struct`. The only time I think that makes a lot of sense is if you are writing a compiler and your struct actually represent a struct in whatever you are compiling :)
</blockquote></details>

Removed.
___
*[nftnl/src/expr/meta.rs, line 95 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-em200fgmbaErxIVc:-M1-vU477wO9KLiBVOfo:b-v63oyh) ([raw file](https://github.com/mullvad/nftnl-rs/blob/744a4282d3fd6d3df7ba4351e6047c0e5598bb04/nftnl/src/expr/meta.rs#L95)):*

Is it *only* mark that can/should be used with `set_value: true`? Then why not have:
```rust
enum Meta {
    ....
    Mark(bool),
    ....
}
```
Instead of creating a new struct that wraps everything.

A goal is that this library should be possible to use cleanly without using the macros. The macros are helpers, but it should be fully possible (and good looking) to use the types directly.
</blockquote></details>

Done. There are a few other options (https://wiki.nftables.org/wiki-nftables/index.php/Setting_packet_metainformation) that can be set, but this still seems simpler.
___
*[nftnl/src/expr/verdict.rs, line 9 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-hQ0X1CYwm5bS1eO2:-M1-yEzj8iOAkWz9mmzF:b-896fix) ([raw file](https://github.com/mullvad/nftnl-rs/blob/744a4282d3fd6d3df7ba4351e6047c0e5598bb04/nftnl/src/expr/verdict.rs#L9)):*

I'm thinking we should maybe keep this type in `immediate.rs`. So we still have a 1:1 mapping between `libnftnl` expressions and Rust modules. So one can look in a single place to see how we implement the `immediate` expression.
</blockquote></details>

Done.

faern (changes requested):
> Reviewed 1 of 5 files at r1, 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/24)* status: all files reviewed, 6 unresolved discussions (waiting on @dlon)

___
*[nftnl/src/expr/ct.rs, line 16 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-fd4rBWkISRbJy9HB:-M10s4mVD_uYkiPSm79P:bp6nm32) ([raw file](https://github.com/mullvad/nftnl-rs/blob/744a4282d3fd6d3df7ba4351e6047c0e5598bb04/nftnl/src/expr/ct.rs#L16)):*

Done. Only `ct mark` can be set, I believe.
</blockquote></details>

This is a great improvement :+1: However I have the same follow-up concern here as with `Meta`: It's hard to interpret what `Mark(true)` vs `Mark(false)` actually do. Implementing them with a named member or as separate enum variants will greatly help with that. And maybe an ounce of documentation.
___
*[nftnl/src/expr/ct.rs, line 51 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M10s_CvEdlfxsfHDfZl:-M10s_CvEdlfxsfHDfZm:b32ne1i) ([raw file](https://github.com/mullvad/nftnl-rs/blob/a94eeb6185b44457ee2a06f665fbc9e274ccd700/nftnl/src/expr/ct.rs#L51)):*
> ```Rust
>         $crate::expr::Conntrack::State
>     };
>     (mark set) => {
> ```

I know this can be perceived as a nitpick. But having the macro variants written in the same order as the enum variants can have a great impact on readability IMO.
___
*[nftnl/src/expr/immediate.rs, line 7 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M10sta99pYIRW9mtiAz:-M10sta99pYIRW9mtiB-:b-wvboym) ([raw file](https://github.com/mullvad/nftnl-rs/blob/a94eeb6185b44457ee2a06f665fbc9e274ccd700/nftnl/src/expr/immediate.rs#L7)):*
> ```Rust
> 
> /// An immediate expression.
> /// Verdicts are handled separately by the Verdict type.
> ```

If you just enclose Verdict in hard brackets, `[Verdict]`, it will be parsed as a link. And `rustdoc` will automatically see it's a local type, so it will create a link to the actual type. That will make the documentation easier to navigate.
___
*[nftnl/src/expr/immediate.rs, line 46 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-wNN-8mspSF0PQbs3:-M10tewv0-rcTQ4M5cmD:b7he3f) ([raw file](https://github.com/mullvad/nftnl-rs/blob/a94eeb6185b44457ee2a06f665fbc9e274ccd700/nftnl/src/expr/immediate.rs#L46)):*

The syntax here is `immediate data <value>`. I'm thinking it should be simply `immediate <value>`. That would be more consistent with the netlink instructions:

```
[ immediate reg 1 <value> ]
[ meta set mark with reg 1 ]
```
</blockquote></details>

Currently `Immediate` only have a single variant. Is it likely it will have more? I have not checked what other stuff one can do with an immediate. If not, optimizing the macro to make the normal `Data` case easier to use does make sense, I agree.
___
*[nftnl/src/expr/meta.rs, line 10 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M10VqgWEV7aQPnGVh1f:-M10VqgWEV7aQPnGVh1g:bmr1ghx) ([raw file](https://github.com/mullvad/nftnl-rs/blob/a94eeb6185b44457ee2a06f665fbc9e274ccd700/nftnl/src/expr/meta.rs#L10)):*
> ```Rust
>     Cgroup,
>     /// Packet mark.
>     Mark(bool),
> ```

I find it hard to understand what this does without reading the implementation *and* a significant amount about `libnftnl`. A potential solution is two completely separate enum variants. `Mark` and `SetMark`. Since they do significantly different things. And they would need to be properly documented to show their difference.

If we want to keep it a single variant it can also be named with `Mark { set: bool },` to make it more clear, because then any user of this variant will type `Meta::Mark(set: true)` which is more understandable than just `Meta::Mark(true)`. But I think the separate variants way is the most understandable?
___
*[nftnl/src/expr/meta.rs, line 46 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M10W-949Mc7wdOoLdw4:-M10W-9577aduIfreWis:bv7kcxc) ([raw file](https://github.com/mullvad/nftnl-rs/blob/a94eeb6185b44457ee2a06f665fbc9e274ccd700/nftnl/src/expr/meta.rs#L46)):*
> ```Rust
>             IifType => libc::NFT_META_IIFTYPE as u32,
>             OifType => libc::NFT_META_OIFTYPE as u32,
>             Mark(_) => libc::NFT_META_MARK as u32,
> ```

Can we please keep the order of the match arms the same as the order the variants are declared in in the enum? That is IMO important for readability.

Regarding the sort order in the enum. I think I have used the same order there as the C header files used, but I can't remember exactly.

Same goes for order of macro variants.

Consistency is key :)

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/24)* status: 1 of 4 files reviewed, 6 unresolved discussions (waiting on @faern)

___
*[nftnl/src/expr/ct.rs, line 16 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-fd4rBWkISRbJy9HB:-M11CeC0276h7t1BNblJ:bk1veid) ([raw file](https://github.com/mullvad/nftnl-rs/blob/744a4282d3fd6d3df7ba4351e6047c0e5598bb04/nftnl/src/expr/ct.rs#L16)):*

This is a great improvement :+1: However I have the same follow-up concern here as with `Meta`: It's hard to interpret what `Mark(true)` vs `Mark(false)` actually do. Implementing them with a named member or as separate enum variants will greatly help with that. And maybe an ounce of documentation.
</blockquote></details>

Updated. It crossed my mind too, but I guess I figured it was clear enough.
___
*[nftnl/src/expr/ct.rs, line 51 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M10s_CvEdlfxsfHDfZl:-M11Cu-d1ALbgDQAbBNP:b-d88gch) ([raw file](https://github.com/mullvad/nftnl-rs/blob/a94eeb6185b44457ee2a06f665fbc9e274ccd700/nftnl/src/expr/ct.rs#L51)):*

I know this can be perceived as a nitpick. But having the macro variants written in the same order as the enum variants can have a great impact on readability IMO.
</blockquote></details>

Updated to use the same order as in the C header.
___
*[nftnl/src/expr/immediate.rs, line 7 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M10sta99pYIRW9mtiAz:-M11D4272iOUo16w3Wpl:b-896fix) ([raw file](https://github.com/mullvad/nftnl-rs/blob/a94eeb6185b44457ee2a06f665fbc9e274ccd700/nftnl/src/expr/immediate.rs#L7)):*

If you just enclose Verdict in hard brackets, `[Verdict]`, it will be parsed as a link. And `rustdoc` will automatically see it's a local type, so it will create a link to the actual type. That will make the documentation easier to navigate.
</blockquote></details>

Done.
___
*[nftnl/src/expr/immediate.rs, line 46 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M1-wNN-8mspSF0PQbs3:-M11E8Fg2r4cJKQU5CKb:b-bulfli) ([raw file](https://github.com/mullvad/nftnl-rs/blob/a94eeb6185b44457ee2a06f665fbc9e274ccd700/nftnl/src/expr/immediate.rs#L46)):*

Currently `Immediate` only have a single variant. Is it likely it will have more? I have not checked what other stuff one can do with an immediate. If not, optimizing the macro to make the normal `Data` case easier to use does make sense, I agree.
</blockquote></details>

I'm not aware of any other use cases, but I'm not certain. Leaving `data` as is but simplifying the code somewhat.
___
*[nftnl/src/expr/meta.rs, line 10 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M10VqgWEV7aQPnGVh1f:-M11DrZU2Ycs4S7tpiVg:bp1x6u4) ([raw file](https://github.com/mullvad/nftnl-rs/blob/a94eeb6185b44457ee2a06f665fbc9e274ccd700/nftnl/src/expr/meta.rs#L10)):*

I find it hard to understand what this does without reading the implementation *and* a significant amount about `libnftnl`. A potential solution is two completely separate enum variants. `Mark` and `SetMark`. Since they do significantly different things. And they would need to be properly documented to show their difference.

If we want to keep it a single variant it can also be named with `Mark { set: bool },` to make it more clear, because then any user of this variant will type `Meta::Mark(set: true)` which is more understandable than just `Meta::Mark(true)`. But I think the separate variants way is the most understandable?
</blockquote></details>

Done: using `Mark { set: bool }`.
___
*[nftnl/src/expr/meta.rs, line 46 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M10W-949Mc7wdOoLdw4:-M11EuUaDmrpZRBATnZN:bf1mw0u) ([raw file](https://github.com/mullvad/nftnl-rs/blob/a94eeb6185b44457ee2a06f665fbc9e274ccd700/nftnl/src/expr/meta.rs#L46)):*

Can we please keep the order of the match arms the same as the order the variants are declared in in the enum? That is IMO important for readability.

Regarding the sort order in the enum. I think I have used the same order there as the C header files used, but I can't remember exactly.

Same goes for order of macro variants.

Consistency is key :)
</blockquote></details>

Good. I've updated it to consistently use the same order as the C header.

faern (approved):
:lgtm:

> Reviewed 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-:-M14tofw1jEucMtgujfC:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[nftnl/src/expr/meta.rs, line 46 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/24#-M10W-949Mc7wdOoLdw4:-M14te-96SIs1r8sVxbc:bb6mt9b) ([raw file](https://github.com/mullvad/nftnl-rs/blob/a94eeb6185b44457ee2a06f665fbc9e274ccd700/nftnl/src/expr/meta.rs#L46)):*

Good. I've updated it to consistently use the same order as the C header.
</blockquote></details>

:+1:

PR #25, opened 2020-02-27
Add base chain types

dlon:
In order to set packet marks, the base chain type must be set to [`route`](https://wiki.nftables.org/wiki-nftables/index.php/Configuring_chains) (rather than `filter`). This is done using `chain.set_base_type(type)`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/25)

### DIFF
--- a/nftnl/src/chain.rs
+++ b/nftnl/src/chain.rs
@@ -1,6 +1,6 @@
 use crate::{MsgType, Table};
 use libc;
-use nftnl_sys::{self as sys, libc::c_void};
+use nftnl_sys::{self as sys, libc::{c_char, c_void}};
 use std::ffi::CStr;
 
 
@@ -31,6 +31,30 @@ pub enum Policy {
     Accept = libc::NF_ACCEPT as u32,
     /// Drop the packet.
     Drop = libc::NF_DROP as u32,
+}
+
+/// Base chain type.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+pub enum ChainType {
+    /// Used to filter packets.
+    /// Supported protocols: ip, ip6, inet, arp, and bridge tables.
+    Filter,
+    /// Used to reroute packets if IP headers or packet marks are modified.
+    /// Supported protocols: ip, and ip6 tables.
+    Route,
+    /// Used to perform NAT.
+    /// Supported protocols: ip, and ip6 tables.
+    Nat,
+}
+
+impl ChainType {
+    fn as_c_str(&self) -> &'static [u8] {
+        match *self {
+            ChainType::Filter => b"filter\0",
+            ChainType::Route => b"route\0",
+            ChainType::Nat => b"nat\0",
+        }
+    }
 }
 
 /// Abstraction of a `nftnl_chain`. Chains reside inside [`Table`]s and they hold [`Rule`]s.
@@ -74,6 +98,18 @@ impl<'a> Chain<'a> {
         unsafe {
             sys::nftnl_chain_set_u32(self.chain, sys::NFTNL_CHAIN_HOOKNUM as u16, hook as u32);
             sys::nftnl_chain_set_u32(self.chain, sys::NFTNL_CHAIN_PRIO as u16, priority);
+        }
+    }
+
+    /// Set the type of a base chain. This only applies if the chain has been registered
+    /// with a hook by calling `set_hook`.
+    pub fn set_type(&mut self, chain_type: ChainType) {
+        unsafe {
+            sys::nftnl_chain_set_str(
+                self.chain,
+                sys::NFTNL_CHAIN_TYPE as u16,
+                chain_type.as_c_str().as_ptr() as *const c_char
+            );
         }
     }
 
--- a/nftnl/src/lib.rs
+++ b/nftnl/src/lib.rs
@@ -62,7 +62,7 @@ pub mod table;
 pub use table::Table;
 
 mod chain;
-pub use chain::{Chain, Hook, Policy, Priority};
+pub use chain::{Chain, ChainType, Hook, Policy, Priority};
 
 mod rule;
 pub use rule::Rule;


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/25)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[nftnl/src/chain.rs, line 107 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/25#-M16KuTsDj9Wb_e8twFS:-M16KuTsDj9Wb_e8twFT:b2gyepw) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4fe418db1ec7bd5251a6f0ba0806c50636ffab9b/nftnl/src/chain.rs#L107)):*
> ```Rust
>     /// Set the type of a base chain. This only applies if the chain is registered
>     /// with a hook by calling `set_hook`.
>     pub fn set_base_type(&mut self, chain_type: ChainType) {
> ```

It would make sense to add this parameter to `set_hook`. Is that a bad idea since it would break the existing interface?
___
*[nftnl/src/chain.rs, line 112 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/25#-M16Kben2ElhT5z2r31n:-M16Kben2ElhT5z2r31o:b-dei98l) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4fe418db1ec7bd5251a6f0ba0806c50636ffab9b/nftnl/src/chain.rs#L112)):*
> ```Rust
>                 self.chain,
>                 sys::NFTNL_CHAIN_TYPE as u16,
>                 CString::new(chain_type.as_str()).unwrap().as_ptr()
> ```

These `CString`s could be static/const rather than allocated each time, I assume?

faern (changes requested):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/25)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[nftnl/src/chain.rs, line 38 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/25#-M16QLm78qFfl_1-ScMc:-M16QLm78qFfl_1-ScMd:b-bqlumi) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4fe418db1ec7bd5251a6f0ba0806c50636ffab9b/nftnl/src/chain.rs#L38)):*
> ```Rust
> /// Base chain type.
> #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
> #[repr(u32)]
> ```

Why does it need a repr at all? I can't see anywhere where the variants are used as their integer representations, nor are they even assigned specific values right now :thinking: 
___
*[nftnl/src/chain.rs, line 41 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/25#-M16QUsE6-7ereK-Pz36:-M16QUsE6-7ereK-Pz37:bd9o2rr) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4fe418db1ec7bd5251a6f0ba0806c50636ffab9b/nftnl/src/chain.rs#L41)):*
> ```Rust
> pub enum ChainType {
>     /// Used to filter packets.
>     /// Supported protocols: ip, ip6, inet, arp, and bridge tables.
> ```

Awesome that you added docs :+1:
___
*[nftnl/src/chain.rs, line 107 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/25#-M16KuTsDj9Wb_e8twFS:-M16Pj_m2ScC0MIcvxou:bkfc8d9) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4fe418db1ec7bd5251a6f0ba0806c50636ffab9b/nftnl/src/chain.rs#L107)):*

It would make sense to add this parameter to `set_hook`. Is that a bad idea since it would break the existing interface?
</blockquote></details>

Breaking the API is fine. It's a `0.x` crate anyway and not that many users (that we know of). And it's more important that the API is nice, ergonomic and understandable than API stability at this point IMO.
___
*[nftnl/src/chain.rs, line 112 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/25#-M16Kben2ElhT5z2r31n:-M16OQh635KZX41SOlVy:bjx2ok5) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4fe418db1ec7bd5251a6f0ba0806c50636ffab9b/nftnl/src/chain.rs#L112)):*

These `CString`s could be static/const rather than allocated each time, I assume?
</blockquote></details>

Yes. There is a lot of converting here that I hope should not be needed. Try something like:
```rust
impl ChainType {
    fn as_c_str(&self) -> &'static [u8] {
        match self {
            ChainType::A => b"aaa\0",
            ChainType::B => b"bbb\0",
        }
    }
}
```
And then avoid using `CString` and `unwrap()`s altogether?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/25)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[nftnl/src/chain.rs, line 38 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/25#-M16QLm78qFfl_1-ScMc:-M16UlZ5CWI3uRvVQ2tm:bsabugo) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4fe418db1ec7bd5251a6f0ba0806c50636ffab9b/nftnl/src/chain.rs#L38)):*

Why does it need a repr at all? I can't see anywhere where the variants are used as their integer representations, nor are they even assigned specific values right now :thinking: 
</blockquote></details>

Fixed. Initially, I had forgotten that this setting was a string.
___
*[nftnl/src/chain.rs, line 107 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/25#-M16KuTsDj9Wb_e8twFS:-M16V027DmlxRQj2eXqB:banyomk) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4fe418db1ec7bd5251a6f0ba0806c50636ffab9b/nftnl/src/chain.rs#L107)):*

Breaking the API is fine. It's a `0.x` crate anyway and not that many users (that we know of). And it's more important that the API is nice, ergonomic and understandable than API stability at this point IMO.
</blockquote></details>

The type is a property of the (base) chain rather than the hook itself. Leaving it separate but renaming it to `set_type`.
___
*[nftnl/src/chain.rs, line 112 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/25#-M16Kben2ElhT5z2r31n:-M16WX0v7KGZF14mvToA:b-9755fa) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4fe418db1ec7bd5251a6f0ba0806c50636ffab9b/nftnl/src/chain.rs#L112)):*

Yes. There is a lot of converting here that I hope should not be needed. Try something like:
```rust
impl ChainType {
    fn as_c_str(&self) -> &'static [u8] {
        match self {
            ChainType::A => b"aaa\0",
            ChainType::B => b"bbb\0",
        }
    }
}
```
And then avoid using `CString` and `unwrap()`s altogether?
</blockquote></details>

Thank you.

faern (approved):
:lgtm:

> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/25#-:-M1zeib1BOOjIjimbct4:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #28, opened 2020-03-30
Fix flags and attributes

dlon:
One issue with older kernels stemmed from `NFTNL_CHAIN_FAMILY` not being set for chains. Some flags were updated as well.

Unrelatedly, add `Chain::dump` for producing a debug string representation of a chain.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/28)

### DIFF
--- a/nftnl/src/chain.rs
+++ b/nftnl/src/chain.rs
@@ -1,7 +1,7 @@
 use crate::{MsgType, Table};
 use libc;
 use nftnl_sys::{self as sys, libc::{c_char, c_void}};
-use std::ffi::CStr;
+use std::ffi::{CStr, CString};
 
 
 pub type Priority = i32;
@@ -116,6 +116,22 @@ impl<'a> Chain<'a> {
                 chain_type.as_c_str().as_ptr() as *const c_char
             );
         }
+    }
+
+    /// Return a string representation of the chain.
+    pub fn dump(&self) -> CString {
+        let mut buffer: [u8; 4096] = [0; 4096];
+        unsafe {
+            sys::nftnl_chain_snprintf(
+                buffer.as_mut_ptr() as *mut i8,
+                buffer.len(),
+                self.chain,
+                0,
+                0,
+            );
+        }
+        let null_index = buffer.iter().position(|ch| *ch == b'\0').unwrap();
+        CString::new(&buffer[0..null_index]).unwrap()
     }
 
     /// Sets the default policy for this chain. That means what action netfilter will apply to
--- a/nftnl/src/chain.rs
+++ b/nftnl/src/chain.rs
@@ -166,11 +166,15 @@ unsafe impl<'a> crate::NlMsg for Chain<'a> {
             MsgType::Add => libc::NFT_MSG_NEWCHAIN,
             MsgType::Del => libc::NFT_MSG_DELCHAIN,
         };
+        let flags: u16 = match msg_type {
+            MsgType::Add => (libc::NLM_F_ACK | libc::NLM_F_CREATE) as u16,
+            MsgType::Del => libc::NLM_F_ACK as u16,
+        };
         let header = sys::nftnl_nlmsg_build_hdr(
             buf as *mut i8,
             raw_msg_type as u16,
             self.table.get_family() as u16,
-            libc::NLM_F_ACK as u16,
+            flags,
             seq,
         );
         sys::nftnl_chain_nlmsg_build_payload(header, self.chain);
--- a/nftnl/src/chain.rs
+++ b/nftnl/src/chain.rs
@@ -4,7 +4,7 @@ use nftnl_sys::{self as sys, libc::{c_char, c_void}};
 use std::ffi::CStr;
 
 
-pub type Priority = u32;
+pub type Priority = i32;
 
 /// The netfilter event hooks a chain can register for.
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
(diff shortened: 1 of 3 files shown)


faern (changes requested):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/28)* status: all files reviewed, 2 unresolved discussions (waiting on @dlon)

___
*[nftnl/src/chain.rs, line 122 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/28#-M3kbCnE0dfwg9arjMUI:-M3kbCnF6Hc0St5wCHDY:b-ltgtw1) ([raw file](https://github.com/mullvad/nftnl-rs/blob/07106c6d4c212a99954e35288e8fbc171fc85fbb/nftnl/src/chain.rs#L122)):*
> ```Rust
> 
>     /// Return a string representation of the chain.
>     pub fn dump(&self) -> CString {
> ```

It's a good idea to use the debug string functionality of `libnftnl`. But `nftnl` being a high level wrapper around `libnftnl` it's supposed to be Rust idiomatic rather than low level. As such I think it would be more correct to implement the `Debug` trait than to return a C string. So convert this method into an `impl fmt::Debug for Chain`?

I think it can use lossy conversion to avoid errors. It's after all just a debug representation.
___
*[nftnl/src/chain.rs, line 134 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/28#-M3ke40O4QVgjkVG3qPs:-M3ke40P-yCNM5NyHtGW:b3jxv2m) ([raw file](https://github.com/mullvad/nftnl-rs/blob/07106c6d4c212a99954e35288e8fbc171fc85fbb/nftnl/src/chain.rs#L134)):*
> ```Rust
>         }
>         let null_index = buffer.iter().position(|ch| *ch == b'\0').unwrap();
>         CString::new(&buffer[0..null_index]).unwrap()
> ```

Instead of manually finding the null byte you can use `CStr::from_ptr` (https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.from_ptr) that does that automatically. And then from the `CStr` to something that can be written in Rust: `CStr::to_string_lossy`.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/28)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[nftnl/src/chain.rs, line 129 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/28#-M3khhdT11c5ISHyIkDc:-M3khhdT11c5ISHyIkDd:b-fa97su) ([raw file](https://github.com/mullvad/nftnl-rs/blob/07106c6d4c212a99954e35288e8fbc171fc85fbb/nftnl/src/chain.rs#L129)):*
> ```Rust
>                 buffer.len(),
>                 self.chain,
>                 0,
> ```

This is the type argument. We should probably use the `nftnl_sys::NFTNL_OUTPUT_DEFAULT` constant instead of hardcoding this to an integer literal.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/28)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[nftnl/src/chain.rs, line 122 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/28#-M3kbCnE0dfwg9arjMUI:-M3pA3GX9DXTkfG4Nd0M:b-896fix) ([raw file](https://github.com/mullvad/nftnl-rs/blob/07106c6d4c212a99954e35288e8fbc171fc85fbb/nftnl/src/chain.rs#L122)):*

It's a good idea to use the debug string functionality of `libnftnl`. But `nftnl` being a high level wrapper around `libnftnl` it's supposed to be Rust idiomatic rather than low level. As such I think it would be more correct to implement the `Debug` trait than to return a C string. So convert this method into an `impl fmt::Debug for Chain`?

I think it can use lossy conversion to avoid errors. It's after all just a debug representation.
</blockquote></details>

Done.
___
*[nftnl/src/chain.rs, line 129 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/28#-M3khhdT11c5ISHyIkDc:-M3pA4ON8Wg7Zmm9uM9s:b-896fix) ([raw file](https://github.com/mullvad/nftnl-rs/blob/07106c6d4c212a99954e35288e8fbc171fc85fbb/nftnl/src/chain.rs#L129)):*

This is the type argument. We should probably use the `nftnl_sys::NFTNL_OUTPUT_DEFAULT` constant instead of hardcoding this to an integer literal.
</blockquote></details>

Done.
___
*[nftnl/src/chain.rs, line 134 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/28#-M3ke40O4QVgjkVG3qPs:-M3pA50B6VG4D9Juo2vL:b-896fix) ([raw file](https://github.com/mullvad/nftnl-rs/blob/07106c6d4c212a99954e35288e8fbc171fc85fbb/nftnl/src/chain.rs#L134)):*

Instead of manually finding the null byte you can use `CStr::from_ptr` (https://doc.rust-lang.org/stable/std/ffi/struct.CStr.html#method.from_ptr) that does that automatically. And then from the `CStr` to something that can be written in Rust: `CStr::to_string_lossy`.
</blockquote></details>

Done.

faern (approved):
:lgtm:

> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/28#-:-M3pOBU-787exSkJuYBS:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[nftnl/src/chain.rs, line 159 at r4](https://reviewable.io/reviews/mullvad/nftnl-rs/28#-M3pLhFUErcGWluU9BkD:-M3pLhFVB4HgooPBXcla:b-66fsia) ([raw file](https://github.com/mullvad/nftnl-rs/blob/508e1433dd6fa89efda815d7119a015c5178473f/nftnl/src/chain.rs#L159)):*
> ```Rust
>         }
>         let s = unsafe { CStr::from_ptr(buffer.as_ptr() as * const c_char) };
>         write!(fmt, "{:?}", s)
> ```

Great! :tada: 

Just nitpick:
I think clippy usually takes note of format strings that are just "{}" or "{:?}". They can be simplified to just something like `s.fmt(fmt)`.

This will also make our debug impl quote the `Chain` string with `"`s. Since the debug impl of `CStr` does that. I don't think that's a blocker. And I'm not sure what the best way around it would be either so probably keep that as is.

PR #32, opened 2020-05-18
Add Reject verdict

dlon:
Responds with an ICMP or TCP RST packet instead of simply dropping packets like `Verdict::Drop`.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/32)


dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/32)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[nftnl/src/expr/verdict.rs, line 118 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7bw-lO8sV693dO39rh:-M7bw-lO8sV693dO39ri:b6o143a) ([raw file](https://github.com/mullvad/nftnl-rs/blob/444aeab6eaa8166ad3de3eecdfb1834ac9a15091/nftnl/src/expr/verdict.rs#L118)):*
> ```Rust
>         $crate::expr::Verdict::Accept
>     };
>     (reject) => {
> ```

Perhaps an optional ICMP code should be added here?

faern (changes requested):
> Reviewed 3 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/32)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @dlon and @faern)

___
*[nftnl/src/expr/verdict.rs, line 3 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7by7zZEj4v2c-kzW8f:-M7by7z_5-ZP3I2VyNVc:btdetfp) ([raw file](https://github.com/mullvad/nftnl-rs/blob/444aeab6eaa8166ad3de3eecdfb1834ac9a15091/nftnl/src/expr/verdict.rs#L3)):*
> ```Rust
> use super::Expression;
> use libc;
> use nftnl_sys::{self as sys, libc::c_char};
> ```

Why does this file `use` libc from two places? :thinking: I know part of this was copied over from another file, but this seems like a leftover or code smell. We should be able to have only one libc.
___
*[nftnl/src/expr/verdict.rs, line 14 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7byJly4SxrB00Z2g_I:-M7byJly4SxrB00Z2g_J:bq7jpg6) ([raw file](https://github.com/mullvad/nftnl-rs/blob/444aeab6eaa8166ad3de3eecdfb1834ac9a15091/nftnl/src/expr/verdict.rs#L14)):*
> ```Rust
>     /// Accept the packet and let it pass.
>     Accept,
>     /// Accept the packet and return an ICMP message.
> ```

"Accept the packet ..."? No it blocks the packet.
___
*[nftnl/src/expr/verdict.rs, line 79 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7cI0ob1_FUaVjhHwkx:-M7cI0ocC2uEvIAP5Cxd:b-my4f3d) ([raw file](https://github.com/mullvad/nftnl-rs/blob/444aeab6eaa8166ad3de3eecdfb1834ac9a15091/nftnl/src/expr/verdict.rs#L79)):*
> ```Rust
>     fn to_expr(&self) -> *mut sys::nftnl_expr {
>         if let Some(immediate_const) = self.immediate_const() {
>             return unsafe { self.immediate_to_expr(immediate_const) };
> ```

I feel the flow here is a bit strange. We check if the local enum represents an immediate. If so we return early with one specific `to_expr` implementation. Otherwise we go back to matching on the enum *again* (we already did this in `immediate_const`).

Trying to always eliminate unreachable branches often creates nicer code that is easier to reason about IMO. How about having two methods on the type:
```rust
/// Exactly what you already have in immediate_to_expr. I just named it a bit differently
fn to_immediate_expr(verdict: i32) -> *mut sys::nftnl_expr {...}

fn to_reject_expr() -> *mut sys::nftnl_expr {...}
```
And then remove `immediate_const` and then in this trait implementation have:
```rust
fn to_expr(&self -> *mut sys::nftnl_expr {
    match *self {
        Verdict::Drop => self.to_immediate_expr(libc::NF_DROP),
        .....
        Verdict::Reject(icmp_code) => self.to_reject_expr(icmp_code),
    }
}
```

That way you get away with no unreachable branches, fewer early returns and IMO a clearer flow from `to_expr` down to the two different implementations of it. The example also has the `icmp_code`. Not sure if it's worth adding that now. Not sure how hard it would be. But if you add it I wanted to showcase approximately how it could be.
___
*[nftnl/src/expr/verdict.rs, line 118 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7bw-lO8sV693dO39rh:-M7byitj4Ny-zhQFRDLY:b-3chz4e) ([raw file](https://github.com/mullvad/nftnl-rs/blob/444aeab6eaa8166ad3de3eecdfb1834ac9a15091/nftnl/src/expr/verdict.rs#L118)):*

Perhaps an optional ICMP code should be added here?
</blockquote></details>

If it's easy to add, please do. Makes the library more versatile and allows us to much easier change the ICMP code later in the daemon.

faern (changes requested):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/32)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @dlon and @faern)

___
*[nftnl/src/expr/verdict.rs, line 79 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7cI0ob1_FUaVjhHwkx:-M7cO559-vBBW9d7YuyC:b-j52647) ([raw file](https://github.com/mullvad/nftnl-rs/blob/444aeab6eaa8166ad3de3eecdfb1834ac9a15091/nftnl/src/expr/verdict.rs#L79)):*

I feel the flow here is a bit strange. We check if the local enum represents an immediate. If so we return early with one specific `to_expr` implementation. Otherwise we go back to matching on the enum *again* (we already did this in `immediate_const`).

Trying to always eliminate unreachable branches often creates nicer code that is easier to reason about IMO. How about having two methods on the type:
```rust
/// Exactly what you already have in immediate_to_expr. I just named it a bit differently
fn to_immediate_expr(verdict: i32) -> *mut sys::nftnl_expr {...}

fn to_reject_expr() -> *mut sys::nftnl_expr {...}
```
And then remove `immediate_const` and then in this trait implementation have:
```rust
fn to_expr(&self -> *mut sys::nftnl_expr {
    match *self {
        Verdict::Drop => self.to_immediate_expr(libc::NF_DROP),
        .....
        Verdict::Reject(icmp_code) => self.to_reject_expr(icmp_code),
    }
}
```

That way you get away with no unreachable branches, fewer early returns and IMO a clearer flow from `to_expr` down to the two different implementations of it. The example also has the `icmp_code`. Not sure if it's worth adding that now. Not sure how hard it would be. But if you add it I wanted to showcase approximately how it could be.
</blockquote></details>

To avoid a lot of `self.to_immediate_expr` repetitions one can also design it like:
```rust
fn to_expr(&self -> *mut sys::nftnl_expr {
    let verdict_const = match *self {
        Verdict::Drop => libc::NF_DROP,
        .....
        Verdict::Reject(icmp_code) => return self.to_reject_expr(icmp_code),
    };
    self.to_immediate_expr(verdict_const)
}
```
It brings back one early return but still eliminated any unreachable execution paths and reduces duplication of the method name a lot.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/32)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[nftnl/src/expr/verdict.rs, line 3 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7by7zZEj4v2c-kzW8f:-M7g9hA47hpzIvoN38Ck:b-896fix) ([raw file](https://github.com/mullvad/nftnl-rs/blob/444aeab6eaa8166ad3de3eecdfb1834ac9a15091/nftnl/src/expr/verdict.rs#L3)):*

Why does this file `use` libc from two places? :thinking: I know part of this was copied over from another file, but this seems like a leftover or code smell. We should be able to have only one libc.
</blockquote></details>

Done.
___
*[nftnl/src/expr/verdict.rs, line 14 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7byJly4SxrB00Z2g_I:-M7g9iiJBbqJ5YsnebRo:bt2vr1b) ([raw file](https://github.com/mullvad/nftnl-rs/blob/444aeab6eaa8166ad3de3eecdfb1834ac9a15091/nftnl/src/expr/verdict.rs#L14)):*

"Accept the packet ..."? No it blocks the packet.
</blockquote></details>

Fixed.
___
*[nftnl/src/expr/verdict.rs, line 79 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7cI0ob1_FUaVjhHwkx:-M7g9oBYBkdteAkGoy--:b-uy34jw) ([raw file](https://github.com/mullvad/nftnl-rs/blob/444aeab6eaa8166ad3de3eecdfb1834ac9a15091/nftnl/src/expr/verdict.rs#L79)):*

To avoid a lot of `self.to_immediate_expr` repetitions one can also design it like:
```rust
fn to_expr(&self -> *mut sys::nftnl_expr {
    let verdict_const = match *self {
        Verdict::Drop => libc::NF_DROP,
        .....
        Verdict::Reject(icmp_code) => return self.to_reject_expr(icmp_code),
    };
    self.to_immediate_expr(verdict_const)
}
```
It brings back one early return but still eliminated any unreachable execution paths and reduces duplication of the method name a lot.
</blockquote></details>

That's a lot better. Done.

faern (changes requested):
> Reviewed 1 of 4 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/32)* status: all files reviewed, 3 unresolved discussions (waiting on @dlon)

___
*[nftnl/src/expr/verdict.rs, line 18 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7gXJL-7glDYScXuYnh:-M7gXJL-7glDYScXuYni:bajnng1) ([raw file](https://github.com/mullvad/nftnl-rs/blob/3804a952d2c1773545ef558d503ffd284c22782f/nftnl/src/expr/verdict.rs#L18)):*
> ```Rust
>         /// `NFT_REJECT_ICMP_UNREACH`: return an ICMP unreachable packet
>         /// `NFT_REJECT_TCP_RST`: reject using TCP RST
>         /// `NFT_REJECT_ICMPX_UNREACH`: ICMP unreachable for inet and bridge
> ```

I suspect this won't render very favorably. Can you check out `cargo doc --open` before we merge this?
___
*[nftnl/src/expr/verdict.rs, line 18 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7gX_qP9UiZ3eTakiX4:-M7gX_qP9UiZ3eTakiX5:b-5rgjup) ([raw file](https://github.com/mullvad/nftnl-rs/blob/3804a952d2c1773545ef558d503ffd284c22782f/nftnl/src/expr/verdict.rs#L18)):*
> ```Rust
>         /// `NFT_REJECT_ICMP_UNREACH`: return an ICMP unreachable packet
>         /// `NFT_REJECT_TCP_RST`: reject using TCP RST
>         /// `NFT_REJECT_ICMPX_UNREACH`: ICMP unreachable for inet and bridge
> ```

This library tries to be a safe and Rust-idiomatic abstraction. Can't you create a Rust enum with the valid reject types? That way it will be nicer to use.
___
*[nftnl/src/expr/verdict.rs, line 25 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7gXkKF0Wf8ZW9h5QkR:-M7gXkKF0Wf8ZW9h5QkS:b-wjq5rb) ([raw file](https://github.com/mullvad/nftnl-rs/blob/3804a952d2c1773545ef558d503ffd284c22782f/nftnl/src/expr/verdict.rs#L25)):*
> ```Rust
>         /// `NFT_REJECT_ICMPX_HOST_UNREACH`, or
>         /// `NFT_REJECT_ICMPX_ADMIN_PROHIBITED`.
>         icmp_code: u8,
> ```

I don't think I understand for which `reject_type` these are valid? All of them? What do they do if reject happens via TCP RST for example?

Same as above, probably nice to make a dedicated enum for the valid values so it can be more Rust idiomatic.

If these are only valid for some `reject_type` you would need to make it possible to only supply them in the correct places. Like the `RejectType` enum can carry an `IcmpRejectCode` member for each variant that allows it?

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/32)* status: 3 of 5 files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[nftnl/src/expr/verdict.rs, line 18 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7gX_qP9UiZ3eTakiX4:-M7j4f713y2chfBzDFFN:b-896fix) ([raw file](https://github.com/mullvad/nftnl-rs/blob/3804a952d2c1773545ef558d503ffd284c22782f/nftnl/src/expr/verdict.rs#L18)):*

This library tries to be a safe and Rust-idiomatic abstraction. Can't you create a Rust enum with the valid reject types? That way it will be nicer to use.
</blockquote></details>

Done.
___
*[nftnl/src/expr/verdict.rs, line 18 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7gXJL-7glDYScXuYnh:-M7lEQZ69rU_Q3aRKqDE:b-896fix) ([raw file](https://github.com/mullvad/nftnl-rs/blob/3804a952d2c1773545ef558d503ffd284c22782f/nftnl/src/expr/verdict.rs#L18)):*

I suspect this won't render very favorably. Can you check out `cargo doc --open` before we merge this?
</blockquote></details>

Done.
___
*[nftnl/src/expr/verdict.rs, line 25 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M7gXkKF0Wf8ZW9h5QkR:-M7j4w4MEkMQiViHio3T:bz8whqb) ([raw file](https://github.com/mullvad/nftnl-rs/blob/3804a952d2c1773545ef558d503ffd284c22782f/nftnl/src/expr/verdict.rs#L25)):*

I don't think I understand for which `reject_type` these are valid? All of them? What do they do if reject happens via TCP RST for example?

Same as above, probably nice to make a dedicated enum for the valid values so it can be more Rust idiomatic.

If these are only valid for some `reject_type` you would need to make it possible to only supply them in the correct places. Like the `RejectType` enum can carry an `IcmpRejectCode` member for each variant that allows it?
</blockquote></details>

Updated. The ICMP code only makes sense for ICMP[X] types.

faern (changes requested):
> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/32)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[nftnl/src/expr/verdict.rs, line 33 at r3](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M8BDKgg83UZRkXQwnBM:-M8BDKgg83UZRkXQwnBN:b-vrar40) ([raw file](https://github.com/mullvad/nftnl-rs/blob/1b720b6e858d315099c403c12e4a64608dabcafd/nftnl/src/expr/verdict.rs#L33)):*
> ```Rust
>     /// Return an ICMP unreachable packet
>     Icmp {
>         family: ProtoFamily,
> ```

We discussed getting rid of this. To not give this responsibility to the user of the library. But instead pass this automatically via the internal code converting exposed types to their underlying libnftnl representations.
___
*[nftnl/src/expr/verdict.rs, line 117 at r3](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M8BBzuN0MZ9pJjDZDrb:-M8BBzuO8-gkUSwhtzln:b-byx7cm) ([raw file](https://github.com/mullvad/nftnl-rs/blob/1b720b6e858d315099c403c12e4a64608dabcafd/nftnl/src/expr/verdict.rs#L117)):*
> ```Rust
>         let reject_code = match reject_type {
>             RejectionType::Icmp { code, ..} => {
>                 code.into()
> ```

I'm not a fan of `.into()` in general. It's very hard to see from the code what type it converts into. If we are going with the `From` impl you have now I think `u8::from(code)` is much more readable. But I think we should be consistent in the library. Other enums representing libc constants have been given the final value directly on the variant and with a `#[repr(type)]`: https://github.com/mullvad/nftnl-rs/blob/master/nftnl/src/chain.rs#L11-L23

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/32)* status: 2 of 14 files reviewed, 2 unresolved discussions (waiting on @faern)

___
*[nftnl/src/expr/verdict.rs, line 33 at r3](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M8BDKgg83UZRkXQwnBM:-M8GET_g7jG2lIWYgcX0:b-896fix) ([raw file](https://github.com/mullvad/nftnl-rs/blob/1b720b6e858d315099c403c12e4a64608dabcafd/nftnl/src/expr/verdict.rs#L33)):*

We discussed getting rid of this. To not give this responsibility to the user of the library. But instead pass this automatically via the internal code converting exposed types to their underlying libnftnl representations.
</blockquote></details>

Done.
___
*[nftnl/src/expr/verdict.rs, line 117 at r3](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M8BBzuN0MZ9pJjDZDrb:-M8G1EjQ6JH_ARd7KcrI:b-896fix) ([raw file](https://github.com/mullvad/nftnl-rs/blob/1b720b6e858d315099c403c12e4a64608dabcafd/nftnl/src/expr/verdict.rs#L117)):*

I'm not a fan of `.into()` in general. It's very hard to see from the code what type it converts into. If we are going with the `From` impl you have now I think `u8::from(code)` is much more readable. But I think we should be consistent in the library. Other enums representing libc constants have been given the final value directly on the variant and with a `#[repr(type)]`: https://github.com/mullvad/nftnl-rs/blob/master/nftnl/src/chain.rs#L11-L23
</blockquote></details>

Done.

faern (approved):
:lgtm:

> Reviewed 10 of 12 files at r4.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-:-M8K-U8eCkuHZXAPOIER:bnfp4nl)* status: 12 of 14 files reviewed, 2 unresolved discussions (waiting on @dlon and @faern)

___
*[nftnl/src/expr/verdict.rs, line 33 at r3](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M8BDKgg83UZRkXQwnBM:-M8K-EXV4x-Ln4FQPhQx:b-ozrqmo) ([raw file](https://github.com/mullvad/nftnl-rs/blob/1b720b6e858d315099c403c12e4a64608dabcafd/nftnl/src/expr/verdict.rs#L33)):*

Done.
</blockquote></details>

A lot better! Nice! Simpler for the user. Fewer strange values they have to get right for it to magically work. :tada: 
___
*[nftnl/src/expr/verdict.rs, line 68 at r4](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M8JyZvL1AZybmwDW2wR:-M8JyZvL1AZybmwDW2wS:b-2myw7i) ([raw file](https://github.com/mullvad/nftnl-rs/blob/bd081996f2708ccd1c31d091cbe6e81c49b75046/nftnl/src/expr/verdict.rs#L68)):*
> ```Rust
>         use libc::*;
>         let value = match code {
>             IcmpCode::NoRoute => NFT_REJECT_ICMPX_NO_ROUTE,
> ```

This match seems to duplicate all the constants. The benefit of assigning the values directly to the variants is that you can now cast the enum directly to a `u8`: `code as u8`. For the other enums where we assign the variants we don't implement conversion to the integer, we just do `as u8` or have some `to_raw(&self)` method.
___
*[nftnl/src/expr/verdict.rs, line 78 at r4](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M8K-2Q_1P7kPoj7YVpK:-M8K-2Q_1P7kPoj7YVpL:b-4v9vgi) ([raw file](https://github.com/mullvad/nftnl-rs/blob/bd081996f2708ccd1c31d091cbe6e81c49b75046/nftnl/src/expr/verdict.rs#L78)):*
> ```Rust
> 
> impl Verdict {
>     unsafe fn immediate_to_expr(&self, immediate_const: i32) -> *mut sys::nftnl_expr {
> ```

I would prefer if the naming was consistent. `to_immediate_expr` would read out nicer and be consistent with `to_reject_expr`. But it's not a public method so I'm not going to block on that.

dlon (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/32)* status: 11 of 14 files reviewed, 1 unresolved discussion (waiting on @faern)

___
*[nftnl/src/expr/verdict.rs, line 68 at r4](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-M8JyZvL1AZybmwDW2wR:-M8KDD103JRAvohZZT0b:b-wt6ius) ([raw file](https://github.com/mullvad/nftnl-rs/blob/bd081996f2708ccd1c31d091cbe6e81c49b75046/nftnl/src/expr/verdict.rs#L68)):*

This match seems to duplicate all the constants. The benefit of assigning the values directly to the variants is that you can now cast the enum directly to a `u8`: `code as u8`. For the other enums where we assign the variants we don't implement conversion to the integer, we just do `as u8` or have some `to_raw(&self)` method.
</blockquote></details>

Done. I forgot to remove this.

faern (approved):
:lgtm:

> Reviewed 2 of 12 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/32#-:-M8KwFJF5M0X2hRg023e:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #34, opened 2020-06-04
Prepare release and make enums non-exhaustive

faern:
`nftnl` just has a small subset of what `libnftnl` supports. Many of the expressions that can go into rules only cover a few of all the possible things an expression can check. And if you add variants to enums it's techincally a breaking change. So in order to not have to deal with this every time, either breaking how versioning is supposed to be done, or making maaany breaking changes, I mark most of the enums that I think will get extended as `#[non_exhaustive]`.

I also bump the version number as I prepare to release now. And this time I treat it as breaking since `#[non_exhaustive]` was not in place in `0.4.0` :shrug:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/34)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 
 ## [Unreleased]
+
+
+
+## [0.5.0] - 2020-06-04
 ### Added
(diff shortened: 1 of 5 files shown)


dlon (approved):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/34)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #35, opened 2020-06-04
Bump libc dependency to the *actual* lowest required

faern:
I tried changing the libc dependency to an exact version (`=0.2.40`) and it failed to build. So we had specified a too low bound. Now it's bumped to the lowest we actually compile with.

To not need to depend on libc twice I just use the one re-exported from the sys crate. Makes it much more clean.

I also update to get `c_void` and `c_char` from the standard library. Since they have been there for quite a while and that is better than relying on a separate libc one. They are effectively the same, but if something is in the standard library that feels more stable.

I also accidentally formatted the code with my stable formatter and was too lazy to revert it. The more crates we can move over to the stable formatter the better IMO also. So I went ahead and removed all unstable formatting options and just went with it.. Like we did on `system-configuration-rs` very recently as well.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/35)

### DIFF
--- a/nftnl-sys/Cargo.toml
+++ b/nftnl-sys/Cargo.toml
@@ -21,7 +21,7 @@ nftnl-1-1-2 = ["nftnl-1-1-1"]
 
 [dependencies]
 cfg-if = "0.1.7"
-libc = "0.2.40"
+libc = "0.2.43"
 
 [build-dependencies]
 cfg-if = "0.1.7"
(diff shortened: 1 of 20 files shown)


dlon (approved):
> Reviewed 20 of 20 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/35)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #36, opened 2020-09-20
Add Nat expressions

faern:
This PR is for a personal toy project of mine. So we don't need this change at Mullvad really. Just want to make that clear :+1: 

This library has no support for source and destination NATing yet. This can be very useful for port forwarding etc. So this PR adds the `Nat` expression type. However! This finally hit the limitations of this library. All expressions we have that store data to a netfilter register have always been hardcoded to use `NFT_REG_1` and all expressions/statements reading from registers have been hardcoded to use the same. This worked fine as long as each data consuming expression only had to read a single value. This is not true for nat expressions which can read up to four registers. There are the `NFTNL_EXPR_NAT_REG_ADDR_MIN`, `NFTNL_EXPR_NAT_REG_ADDR_MAX`, `NFTNL_EXPR_NAT_REG_PROTO_MIN` and `NFTNL_EXPR_NAT_REG_PROTO_MAX` values to set on the nat expression. I currently don't care about supporting ranges of IPs or ports. So I skip the `_MAX` variant. Only setting the `_MIN` variants means it's not a range.

I honestly don't know exactly what the ranges are for. But I guess it's for load balancing. There is very little documentation on this entire library sadly. The man page for `nft` nor the nftables wiki mentions the ranges except that it's a valid syntax to input :shrug: 

Anyway. In order to fix the problem of having an expression that reads from multiple registers, I changed the `Immediate` expression to also take a `Register` enum that controls which register it writes to. This is fine for now. This is of course a breaking change. But I made the macro kind of hide this for now as a workaround. So anyone using the macro won't notice the difference.

## Examples

This is how one can create a port forwarding rule now:
```rust
        let public_port = 1234u16;
        let ip_to_forward_to = Ipv4Addr::new(10, 0, 100, 200);
        let port_to_forward_to = 4321u16;

        let mut rule = Rule::new(&prerouting_chain);

        rule.add_expr(&nft_expr!(meta l4proto));
        rule.add_expr(&nft_expr!(cmp == libc::IPPROTO_TCP as u8));

        rule.add_expr(&nft_expr!(payload tcp dport));
        rule.add_expr(&nft_expr!(cmp == public_port.to_be()));

        rule.add_expr(&Immediate::new(ip_to_forward_to, expr::Register::Reg1));
        rule.add_expr(&Immediate::new(port_to_forward_to.to_be(), expr::Register::Reg2));
        let nat_expr = expr::Nat {
            nat_type: expr::NatType::DNat,
            family: ProtoFamily::Ipv4,
            ip_reg: expr::Register::Reg1,
            port_reg: Some(expr::Register::Reg2),
        };
        rule.add_expr(&nat_expr);
```
It results in a rule like this:
```
tcp dport 1234 dnat to 10.0.100.200:4321
```

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/36)

### DIFF
--- a/nftnl/src/expr/immediate.rs
+++ b/nftnl/src/expr/immediate.rs
@@ -1,5 +1,5 @@
-use super::{Expression, Rule};
-use nftnl_sys::{self as sys, libc};
+use super::{Expression, Register, Rule};
+use nftnl_sys as sys;
 use std::ffi::c_void;
 use std::mem::size_of_val;
 use std::os::raw::c_char;
@@ -9,6 +9,13 @@ use std::os::raw::c_char;
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
 pub struct Immediate<T> {
     pub data: T,
+    pub register: Register,
+}
+
+impl<T> Immediate<T> {
+    pub fn new(data: T, register: Register) -> Self {
+        Self { data, register }
+    }
 }
 
 impl<T> Expression for Immediate<T> {
@@ -21,7 +28,7 @@ impl<T> Expression for Immediate<T> {
             sys::nftnl_expr_set_u32(
                 expr,
                 sys::NFTNL_EXPR_IMM_DREG as u16,
-                libc::NFT_REG_1 as u32,
+                self.register.to_raw(),
             );
 
             sys::nftnl_expr_set(
@@ -39,6 +46,9 @@ impl<T> Expression for Immediate<T> {
 #[macro_export]
 macro_rules! nft_expr_immediate {
     (data $value:expr) => {
-        $crate::expr::Immediate { data: $value }
+        $crate::expr::Immediate {
+            data: $value,
+            register: Register::Reg1,
+        }
     };
 }
--- a/nftnl/src/expr/mod.rs
+++ b/nftnl/src/expr/mod.rs
@@ -4,13 +4,30 @@
 //! [`Rule`]: struct.Rule.html
 
 use super::rule::Rule;
-use nftnl_sys as sys;
+use nftnl_sys::{self as sys, libc};
 
 /// Trait for every safe wrapper of an nftables expression.
 pub trait Expression {
     /// Allocates and returns the low level `nftnl_expr` representation of this expression.
     /// The caller to this method is responsible for freeing the expression.
     fn to_expr(&self, rule: &Rule) -> *mut sys::nftnl_expr;
+}
+
+/// A netfilter data register. The expressions store and read data to and from these
+/// when evaluating rule statements.
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Hash)]
+#[repr(i32)]
+pub enum Register {
+    Reg1 = libc::NFT_REG_1,
+    Reg2 = libc::NFT_REG_2,
+    Reg3 = libc::NFT_REG_3,
+    Reg4 = libc::NFT_REG_4,
+}
+
+impl Register {
+    pub fn to_raw(self) -> u32 {
+        self as u32
+    }
 }
 
 mod bitwise;
--- a/nftnl/src/expr/mod.rs
+++ b/nftnl/src/expr/mod.rs
@@ -54,6 +54,9 @@ pub use self::masquerade::*;
 mod meta;
 pub use self::meta::*;
 
+mod nat;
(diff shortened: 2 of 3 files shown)


dlon (approved):
This is hard to do with the current design, but it might be nice if some expressions could be abstracted away, e.g.
```
rule.add_expression(DPort::new(1234));
rule.add_expression(DNat::new(Ipv4Addr::from(10, 0, 100, 200), 4321));
```
You might do something like
```
impl Expression for DNat {
    fn to_expr(&self, _: &Rule) -> Vec<*mut sys::nftnl_expr> {
        let mut expressions = Vec::new();
        expressions.push(Immediate::new(self.to_ip, Register::Reg1).to_expr().remove(0));
        expressions.push(Immediate::new(self.to_port, Register::Reg2).to_expr().remove(0));
        ...
        expressions
    }
}
```

> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/36#-:-MHjpnk067QTq0zR_r9p:blyvv1r)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[nftnl/src/expr/nat.rs, line 20 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/36#-MHjpG_TD-R0YtYIYHZV:-MHjpG_U9P3-vuVcLafW:bvf70ee) ([raw file](https://github.com/mullvad/nftnl-rs/blob/fcd334cb2acc38933ae680290e1e433bb5e381ea/nftnl/src/expr/nat.rs#L20)):*
> ```Rust
>     pub nat_type: NatType,
>     pub family: ProtoFamily,
>     pub ip_reg: Register,
> ```

Maybe it should be called `ip_register` instead of `ip_reg`?

faern (commented):
I have been thinking about reasonable abstractions as well. That's one way. But also, instead of `to_expr` returning the raw expressions, they could just add themselves (variable amount of actual expressions) directly to the rules, they already get the rule as input argument. Then they don't have to return anything.

I kind of wrote the `DNat` thing you propose above. But I realize it has some limitations. For example see this rule:
```
dnat to numgen random mod 10 map { 0-2 : 192.168.10.100, 3-9 : 192.168.20.200 }
```
The above is an example rule from `man nft` that does load balancing. I have not yet ran this rule through `nft` with a debugger attached. So I don't know exactly how it's implemented. But I assume it starts with some randomization expression that then loads the selected IP into a register and then executes the nat expression. If we have `struct DNat { ip: IpAddr }` we can't express this. We would kind of need something like:
```rust
struct DNat<A: OutputExpression<Type = IpAddr>, P: OutputExpression<Type = u16>> {
    ip: A,
    port: Option<P>
}
```
And have `IpAddr` and `u16` implement `OutputExpression` etc... But this netfilter "language" is pretty complex. So we would likely not cover everything... So what I'm debating with myself is if it's worth trying to create a very advanced abstraction that makes it impossible to create invalid rules, but might make it impossible to create certain special case rules. Or if the library should allow creating any expression in any order, but with the possibility of creating rules that makes no sense.. I honestly don't know really, but the latter sure is easier :sweat_smile: 

> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/36#-:-MHjuj1V6jiDxQikYzvE:b-wzae89)* status: 2 of 3 files reviewed, all discussions resolved (waiting on @dlon)

___
*[nftnl/src/expr/nat.rs, line 20 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/36#-MHjpG_TD-R0YtYIYHZV:-MHjv10x9aJkTwvp8Tr2:b-smrwkr) ([raw file](https://github.com/mullvad/nftnl-rs/blob/fcd334cb2acc38933ae680290e1e433bb5e381ea/nftnl/src/expr/nat.rs#L20)):*

Maybe it should be called `ip_register` instead of `ip_reg`?
</blockquote></details>

Ok. Yes I agree. I'm usually the one saying to not use abbreviations unless really needed. So I should not do that myself. Will fix :+1:

dlon (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/36)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #37, opened 2020-10-06
Make Batch,Iter,Table,Rule&Chain implement Send+Sync

faern:
Implementing `Send` for a type allows it to transition in owned form between threads. Only one thread can own it and use it at any given time. Implementing `Sync` means immutable references to the instance are allowed to transition between threads. This means there can be multiple immutable references to the same instance alive at the same time in multiple threads. But since they are immutable borrows all threads can only "read" from the instance, or call `&self` methods.

A lot of the types in `nftnl` contain raw pointers of types from the underlying `-sys` crate. So by definition Rust is conservative here and assume the types are not `Send` nor `Sync` for safety. It's up to the implementor to prove those traits hold and unsafely implement them.

I have not proven in any way if these types are safe to access (immutably) from multiple threads at once, but I can't imagine why they would not be. All methods that modify the memory should be taking `&mut self` anyway and can as a result not be called in parallel and screw things up anyway. For the same reason `Vec` from the standard library implements both these traits. But I could be wrong of course.

The reason for this PR is that I'm trying to use `nftnl` in an async context in a private project. And requirements on futures are usually that they are `Send + Sync` so they can be moved between worker threads between yield points etc. As a result, this library is not really usable without implementing these traits for most types that are used when modifying the firewall.

I thought this could be usable at Mullvad as well since we move more and more stuff to async, and eventually we might want to do it with the firewall as well?

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/37)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 
 ## [Unreleased]
-
+### Added
+- Implement Send+Sync for Table, Chain, Rule, Batch and Iter (batch iterator)
 
 
 ## [0.5.0] - 2020-06-04
--- a/nftnl/src/batch.rs
+++ b/nftnl/src/batch.rs
@@ -25,6 +25,11 @@ pub struct Batch {
     seq: u32,
 }
 
+// Safety: It should be safe to pass this around and *read* from it
+// from multiple threads
+unsafe impl Send for Batch {}
+unsafe impl Sync for Batch {}
+
 impl Batch {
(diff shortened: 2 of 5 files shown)


faern:
I have verified that I can add and remove firewall rules using the async version of `netlink-sys` with this PR applied to this library.

dlon (approved):
:lgtm:

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/37#-:-MJ1Z_50B2LNiATiIze6:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #38, opened 2020-10-06
Upgrade cfg-if dependency to 1.0

faern:
I just happened to stumble on the fact that `cfg-if` came out with version `1.0.0.` just hours ago. It's a pretty small crate and rather simple macro. So I guess this means they have decided on the final shape and form of that macro. Which is good. Probably means we virtually never have to upgrade it again :)

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/38)


dlon (approved):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/38)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #46, opened 2021-09-28
fixes #45 - Add a log expression with group argument

### LINKED ISSUE #45: How to add a log rule
I haven't managed to add a log rule to a chain. Is there a process I missed ? Or is it not yet implemented ? The [nftables wiki](https://wiki.nftables.org/wiki-nftables/index.php/Logging_traffic) mentions a `filter` expr, that I haven't found in nftnl-rs' source.

lafleurdeboum:
This PR adds a `log` expression, solving #45 . To add a log rule do

```rust
use std::ffi::CString;
use nftnl::{Rule, Chain, nft_expr};
let chain = Chain::new(&CString::new("some-chain-name")?);
let rule = Rule::new(&chain);
rule.add_expr(&nft_expr!(log));
```
And then insert in the chain where you bet useful.

To set the `group` argument to a value between zero and seven, pick a value in the `expr::LogGroup` enum and do :
```rust
use std::ffi::CString;
use nftnl::{Rule, Chain, nft_expr, expr::LogGroup::LogGroupZero};
let chain = Chain::new(&CString::new("some-chain-name")?);
let rule = Rule::new(&chain);
rule.add_expr(&nft_expr!(log group LogGroupZero));
```
As stated in [nftables' wiki](https://wiki.nftables.org/wiki-nftables/index.php/Logging_traffic), setting the `group` argument activates NFLog output, useable by ulogd2, a more advanced logging target than the default. The default is to output to kernel log, which is a pity, because it's often unprotected. ulogd needs a running daemon, and lets you choose output targets and log format.

So this PR is probably not the most urgent thing in Mullvad userland, but can prove useful, and is very compact. Besides, it _is_ very much needed for other applications that would use this library for its quality. `nftnl-rs` rocks, by the way ! :1st_place_medal:

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/46)

### DIFF
--- a/nftnl/src/expr/log.rs
+++ b/nftnl/src/expr/log.rs
@@ -6,10 +6,32 @@ use nftnl_sys::{
 
 /// A log expression.
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
-pub struct Log {}
+pub enum Log {
+    NoGroup,
+    Group(LogGroup),
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+pub enum LogGroup {
+    LogGroupZero,
+    LogGroupOne,
+    LogGroupTwo,
+    LogGroupThree,
+    LogGroupFour,
+    LogGroupFive,
+    LogGroupSix,
+    LogGroupSeven,
+}
 
 
-impl Log {}
+impl Log {
+    pub fn new() -> Self {
+        Log::NoGroup
+    }
+    pub fn new_with_group(group: LogGroup) -> Self {
+        Log::Group(group)
+    }
+}
 
 impl Expression for Log {
     fn to_expr(&self, _rule: &Rule) -> *mut sys::nftnl_expr {
@@ -17,7 +39,16 @@ impl Expression for Log {
             let expr = try_alloc!(sys::nftnl_expr_alloc(
                 b"log\0" as *const _ as *const c_char
             ));
-
+            match self {
+                Log::NoGroup => (),
+                Log::Group(group) => {
+                    sys::nftnl_expr_set_u32(
+                        expr,
+                        sys::NFTNL_EXPR_LOG_GROUP as u16,
+                        *group as u32,
+                    );
+                }
+            }
             expr
         }
     }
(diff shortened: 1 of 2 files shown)


lafleurdeboum:
@faern would you consider commenting this PR ? I believe it can't break anything from any mullvad crate, since it simply adds functionality that was nonexistent until then. Besides, it is (arguably) quite needed to use `nftnl-rs` as a main linux firewall library, since logging is a critical concern for those.

nftnl/src/expr/log.rs
```diff
@@ -0,0 +1,57 @@
...
+    libc::c_char,
+};
+
+/// A log expression.
+pub struct Log {
+    group: Option<LogGroup>,
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+pub enum LogGroup {
```
dlon, line 13: I couldn't find any reason for limiting this to 8. `u16` seems to be the [actual](https://git.netfilter.org/libnftnl/tree/src/expr/log.c) type, so I think it would make sense to use that.

nftnl/src/expr/mod.rs
```diff
@@ -71,6 +74,12 @@ macro_rules! nft_expr {
     (cmp $op:tt $data:expr) => {
         nft_expr_cmp!($op $data)
     };
+    (log group $group:ident) => {
+        nft_expr_log!(group $group)
+    };
+    (log) => {
+        nft_expr_log!()
```
dlon, line 81: Could you call `crate::expr::Log::new(None)` here directly instead?

nftnl/src/expr/log.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl Expression for Log {
+    fn to_expr(&self, _rule: &Rule) -> *mut sys::nftnl_expr {
+        unsafe {
+            let expr = try_alloc!(sys::nftnl_expr_alloc(
+                b"log\0" as *const _ as *const c_char
+            ));
+            if let Some(group) = self.group {
+                sys::nftnl_expr_set_u32(
```
dlon, line 38: This should be `nftnl_expr_set_u16`.

dlon (changes requested):
Thanks for the PR! 👍

faern:
Thanks for wanting to contribute! However, due to inactivity (from both sides I have to say), this is superseded by #56

PR #47, opened 2021-10-08
Support checking conntrack status

joshtriplett:
Add `ct status` to load the conntrack status, and add conntrack status
bitflags.

As an example, the rule `ct status dnat masquerade` can be written as:

```rust
let mut rule_masq = nftnl::Rule::new(&chain_postrouting);
rule_masq.add_expr(&nft_expr!(ct status));
rule_masq.add_expr(&nft_expr!(bitwise mask nftnl::expr::ConntrackStatus::DST_NAT.bits(), xor 0));
rule_masq.add_expr(&nft_expr!(cmp != 0));
rule_masq.add_expr(&nft_expr!(masquerade));
batch.add(&rule_masq, nftnl::MsgType::Add);
```

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/47)

### DIFF
--- a/nftnl/src/expr/ct.rs
+++ b/nftnl/src/expr/ct.rs
@@ -12,8 +12,30 @@ bitflags::bitflags! {
     }
 }
 
+bitflags::bitflags! {
+    pub struct ConntrackStatus: u32 {
+        const EXPECTED = 1;
+        const SEEN_REPLY = 2;
+        const ASSURED = 4;
+        const CONFIRMED = 8;
+        const SRC_NAT = 16;
+        const DST_NAT = 32;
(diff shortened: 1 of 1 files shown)


dlon (commented):
Looks good!

Are all of the statuses defined in `nf_conntrack_common.h` usable with nftables? This wiki hints that only some are: https://wiki.nftables.org/wiki-nftables/index.php/Matching_connection_tracking_stateful_metainformation#ct_status_-_conntrack_status.

If not, we should probably only include those ones.

dlon (approved):
LGTM. This is fine, actually.

PR #48, opened 2021-11-14
Update libc dependency to 0.2.107

pulsastrix:
This PR updates the libc crate dependency to the most recently released version.

Compared to the version that was used previously (0.2.43), the new version contains the changes made in rust-lang/libc#2152 (and released in libc version 0.2.95), which allows nftnl-rs to be used on Linux systems running musl-libc instead of glibc.

For nftnl-rs to work on musl-libc systems, the libc dependency in mullvad/mnl-rs has to be updated as well. I have opened a PR in the mnl-rs repository to do so: mullvad/mnl-rs#11. The version of mnl-rs used in this repository should probably be updated accordingly before merging this PR.

---
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/48)

### DIFF
--- a/nftnl-sys/Cargo.toml
+++ b/nftnl-sys/Cargo.toml
@@ -21,7 +21,7 @@ nftnl-1-1-2 = ["nftnl-1-1-1"]
 
 [dependencies]
 cfg-if = "1.0"
-libc = "0.2.43"
+libc = "0.2.107"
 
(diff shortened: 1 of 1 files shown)


dlon:
0.2.107 should be compatible with the toml file. You don't need to change anything there, just run `cargo update -p libc`.

PR #54, opened 2023-12-03
removes expect statments from nft_set macro

cjmakes:
This commit removes expect statements from the nft_set macro which cause compile errors since the Result was remove in #21 

small repro:

```rust
use nftnl::{nft_set, set::Set, Table, ProtoFamily};
use std::ffi::CString;

fn main() {
   let t = Table::new(&CString::new("test").unwrap(), nftnl::ProtoFamily::Ipv4);
   let s:Set<std::net::Ipv4Addr> = nft_set!(&CString::new("s1").unwrap(), 0, &t, ProtoFamily::Ipv4; [ &[1,1,1,1].into(), ]);
}

```

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/54)

### DIFF
--- a/nftnl/src/set.rs
+++ b/nftnl/src/set.rs
@@ -17,9 +17,9 @@ macro_rules! nft_set {
         nft_set!($name, $id, $table, $family)
     };
     ($name:expr, $id:expr, $table:expr, $family:expr; [ $($value:expr,)* ]) => {{
-        let mut set = nft_set!($name, $id, $table, $family).expect("Set allocation failed");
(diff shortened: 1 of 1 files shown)


faern:
Oops. This is a mistake on our part! It was a mistake to not have any examples or tests at all for sets. We have now added a proper CI to this crate. Could you maybe rebase this PR to have it pass the CI? Best would of course also be if you could add doctests or an example how sets are used. So we get a compile time check that they still work :)

faern:
Thank you for the contribution. We'll look into adding set tests/docs separately.

PR #57, opened 2024-05-29
Add lots of Github Actions CI jobs to check sanity

faern:
This crate has gone without any CI at all for a long time. We have no unit tests in this code currently. But we can and should lots of other stuff:
* Linting
* Formatting
* Git commit message style
* Does it build?
* Does it have known vulnerabilities in dependencies?

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/57)


faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/57)* status: 0 of 13 files reviewed, 1 unresolved discussion

___
*[`.github/workflows/build-and-test.yml` line 71 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/57#-Nz4sVmfDfhXzd6ov6R9:-Nz4sVmfDfhXzd6ov6RA:bots0fw) ([raw file](https://github.com/mullvad/nftnl-rs/blob/92607e8b5638aad2713083445df2525de7652f76/.github/workflows/build-and-test.yml#L71)):*
> ```YAML
> 
>       - name: Compile with minimal versions
>         run: cargo +stable build --all-targets
> ```

Here I tried a slightly different approach than what I merged in https://github.com/mullvad/mnl-rs/pull/14 and https://github.com/mullvad/pfctl-rs/pull/89. I realized doing the `cargo build` step with nightly would be less stable. Nightly causes issues more frequently than stable. Only reason this job needs nightly at all is to do the `-Z minimal-versions` part.

Stable Rust is already available in the Github actions containers. So installing it is basically free anyway. The only reason I explicitly install it above instead of relying on it being installed is that when a new Rust version is shipped, it takes a while before the Github Actions containers are upgraded. By explicitly doing this we make sure we always use the same `stable` toolchain across the different CI jobs.

If we agree this is better I'll submit this improvement to the other two repositories as well.

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/57)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @Serock3)

___
*[`.github/workflows/build-and-test.yml` line 46 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/57#-O-5NFZH2s4PeiH5XnIb:-O-5NFZH2s4PeiH5XnIc:b-efd25b) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4a84e9c6db9d02f50e22fd1207fd1ad1940855b2/.github/workflows/build-and-test.yml#L46)):*
> ```YAML
>       - name: Generate documentation
>         if: matrix.rust == 'stable'
>         run: RUSTDOCFLAGS="--deny warnings" cargo doc
> ```

FYI: The reason I decided to set `RUSTDOCFLAGS` here, but `RUSTFLAGS` is set up in the `env` section of the file is that `RUSTDOCFLAGS` is only relevant for this single `cargo` invocation, while the `RUSTFLAGS` is relevant to set for all the other jobs in this file. If we decide to move the rustdoc check to another workflow file, it would be harder to remember moving `RUSTDOCFLAGS` if it was specified somewhere completely different. Specifying `RUSTFLAGS` globally saves us from duplication, but the same is not true for the docs flag.

Serock3 (approved):
:lgtm:

> Reviewed 6 of 13 files at r1, 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/57#-:-O-5NwCv-Imatj1yTH32:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @faern)

___
*[`.github/workflows/build-and-test.yml` line 71 at r1](https://reviewable.io/reviews/mullvad/nftnl-rs/57#-Nz4sVmfDfhXzd6ov6R9:-O-6aNxcBlCaHcz7Ruxj:b-l4dd3l) ([raw file](https://github.com/mullvad/nftnl-rs/blob/92607e8b5638aad2713083445df2525de7652f76/.github/workflows/build-and-test.yml#L71)):*

Here I tried a slightly different approach than what I merged in https://github.com/mullvad/mnl-rs/pull/14 and https://github.com/mullvad/pfctl-rs/pull/89. I realized doing the `cargo build` step with nightly would be less stable. Nightly causes issues more frequently than stable. Only reason this job needs nightly at all is to do the `-Z minimal-versions` part.

Stable Rust is already available in the Github actions containers. So installing it is basically free anyway. The only reason I explicitly install it above instead of relying on it being installed is that when a new Rust version is shipped, it takes a while before the Github Actions containers are upgraded. By explicitly doing this we make sure we always use the same `stable` toolchain across the different CI jobs.

If we agree this is better I'll submit this improvement to the other two repositories as well.
</blockquote></details>

This seems like a sound argument to me
___
*[`.github/workflows/build-and-test.yml` line 41 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/57#-O-5gxVj6Ociozd1hoin:-O-5gxVj6Ociozd1hoio:b-7bt80m) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4a84e9c6db9d02f50e22fd1207fd1ad1940855b2/.github/workflows/build-and-test.yml#L41)):*
> ```YAML
> 
>       - name: Test
>         run: cargo test
> ```

Should the tests not be run with the `--locked` flag to prevent `cargo test` from rebuilding of a bunch of deps?
___
*[`.github/workflows/build-and-test.yml` line 46 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/57#-O-5NFZH2s4PeiH5XnIb:-O-5gZ0PCblnKV5-gf7J:b-my31f8) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4a84e9c6db9d02f50e22fd1207fd1ad1940855b2/.github/workflows/build-and-test.yml#L46)):*

FYI: The reason I decided to set `RUSTDOCFLAGS` here, but `RUSTFLAGS` is set up in the `env` section of the file is that `RUSTDOCFLAGS` is only relevant for this single `cargo` invocation, while the `RUSTFLAGS` is relevant to set for all the other jobs in this file. If we decide to move the rustdoc check to another workflow file, it would be harder to remember moving `RUSTDOCFLAGS` if it was specified somewhere completely different. Specifying `RUSTFLAGS` globally saves us from duplication, but the same is not true for the docs flag.
</blockquote></details>

Alright
___
*[`.github/workflows/cargo-audit.yml` line 27 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/57#-O-5bFQQB9cHSnFKK72A:-O-5bFQQB9cHSnFKK72B:b3nuf5) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4a84e9c6db9d02f50e22fd1207fd1ad1940855b2/.github/workflows/cargo-audit.yml#L27)):*
> ```YAML
>       # This avoids significant maintenance work that provide no benefits.
>       # We only need to make sure there is any compatible dependency without a known issue
>       - run: cargo update
> ```

Is it necessary to run `cargo update` when we are not providing the `file: Cargo.lock` argument?
___
*[`.github/workflows/linting.yml` line 7 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/57#-O-5enZ893kShUpeEvHr:-O-5enZ893kShUpeEvHs:b-ratqoz) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4a84e9c6db9d02f50e22fd1207fd1ad1940855b2/.github/workflows/linting.yml#L7)):*
> ```YAML
>     paths:
>       - .github/workflows/linting.yml
>       - '**/*.rs'
> ```

I don't suppose we want to re-run clippy on changes to `Cargo.toml`? Update a dependency could cause a clippy lint to trigger from e.g. a deprecation, right?

faern (commented):
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/57)* status: all files reviewed, 3 unresolved discussions (waiting on @Serock3)

___
*[`.github/workflows/linting.yml` line 7 at r2](https://reviewable.io/reviews/mullvad/nftnl-rs/57#-O-5enZ893kShUpeEvHr:-O-6bvT96kaLcHkSfBCM:b559imc) ([raw file](https://github.com/mullvad/nftnl-rs/blob/4a84e9c6db9d02f50e22fd1207fd1ad1940855b2/.github/workflows/linting.yml#L7)):*
<details><summary><i>Previously, Serock3 (Sebastian Holmin) wrote…</i></summary><blockquote>

I don't suppose we want to re-run clippy on changes to `Cargo.toml`? Update a dependency could cause a clippy lint to trigger from e.g. a deprecation, right?
</blockquote></details>

Deprecation is a `rustc` warning, not a `clippy` warning. Right? Those warnings should be emitted by `cargo build`

PR #58, opened 2024-05-29
Remove special errors in examples

faern:
They were not properly used anyway. Just returning these errors would print better messages if there actually are any errors

Cargo was complaining about these. But I had never seen it because we had no CI and I probably did not build them in many years.

Yes, we don't have a CI. I work on one in a different PR. But I did not want that PR to be too large. And this fix is needed to make the CI work. So I opted for doing this first.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/58)


Serock3 (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/58)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #59, opened 2024-05-29
Replace `err_derive` dependency with manual impl of Display + Error

faern:
Building this library on beta or nightly yields the following warning:
```
warning: non-local `impl` definition, `impl` blocks should be written at the same level as their item
  --> nftnl/src/batch.rs:8:10
   |
8  | #[derive(err_derive::Error, Debug)]
   |          ^----------------
   |          |
   |          `Error` is not local
   |          move the `impl` block outside of this constant `_DERIVE_std_error_Error_FOR_NetlinkError`
9  | #[error(display = "Error while communicating with netlink")]
10 | pub struct NetlinkError(());
   |            ------------ `NetlinkError` is not local
```

We only have a single error in this crate, and it's an extremely simple one. Here I just replace this derive macro with a simple manual implementation! Less weird bugs and much more future proof! And fewer dependencies! :partying_face:  This change removes *eleven* dependencies from the dependency tree of this (should be) small library.

Getting rid of these warnings is a prerequisite for getting #57 into a working state.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/59)

### DIFF
--- a/nftnl/Cargo.toml
+++ b/nftnl/Cargo.toml
@@ -20,7 +20,6 @@ nftnl-1-1-2 = ["nftnl-sys/nftnl-1-1-2"]
 
 [dependencies]
 bitflags = "1.0.4"
-err-derive = "0.3.1"
 log = "0.4"
 nftnl-sys = { path = "../nftnl-sys", version = "0.6.1" }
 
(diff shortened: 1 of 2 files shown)


Serock3 (approved):
🙌

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/59#-:-NzTIVe28InljGtZQQp0:b44u44)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #63, opened 2024-06-04
docs: fix docs.rs build without linking

my4ng:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/63)


faern (commented):
Nice! Thank you for wanting to contribute. getting the docs to build on docs.rs would be awesome. Do you have any link or reference explaining this is the recommended/a common way of solving this issue?

faern:
I found this: https://docs.rs/about/builds#detecting-docsrs. So it makes sense! Will merge, thank you.

PR #65, opened 2024-06-22
fix: usage of legacy numerical constant

hirehamir:
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/65)

### DIFF
--- a/nftnl/src/lib.rs
+++ b/nftnl/src/lib.rs
@@ -127,5 +127,5 @@ pub unsafe trait NlMsg {
 /// length (nla_len) is 16 bits, the largest message is a bit larger than
 /// 64 KBytes.
 pub fn nft_nlmsg_maxsize() -> u32 {
-    u32::from(::std::u16::MAX) + unsafe { libc::sysconf(libc::_SC_PAGESIZE) } as u32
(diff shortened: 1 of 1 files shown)


nftnl/src/lib.rs
```diff
@@ -127,5 +127,5 @@ pub unsafe trait NlMsg {
 /// length (nla_len) is 16 bits, the largest message is a bit larger than
 /// 64 KBytes.
 pub fn nft_nlmsg_maxsize() -> u32 {
-    u32::from(::std::u16::MAX) + unsafe { libc::sysconf(libc::_SC_PAGESIZE) } as u32
+    u32::from(u16::MAX) + unsafe { libc::sysconf(libc::_SC_PAGESIZE) } as u32
```
hirehamir, line 130: `std::<integer>::MAX` has been superseded by the associated constant on the type, e.g., `u16::MAX`.

The legacy version could be deprecated in a future version of Rust. Also, this version of the code is probably more readable.

faern:
This still fails with the same type of error, but in an example. I'll merge this and just fix it in the example also.

faern:
Thanks for the contribution :)

hirehamir:
You're welcome!

PR #66, opened 2024-07-18
Upgrade bitflags dependency to latest (2.6.0)

faern:
Just a nice cleanup, adopting latest `bitflags`. Not much changes in practice. One thing is that with bitflags 2 all the traits are not automatically implemented, so one has to do that explicitly. I list all traits that the old enums implemented except `Ord` and `PartialOrd`. I don't think it makes much sense to "order" instances of bitflags? What does that even mean? All the other traits I do think make sense.

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/66)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Upgrade crates to Rust 2021 edition.
 
+### Removed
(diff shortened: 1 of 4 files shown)


Serock3 (approved):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/66)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #67, opened 2024-07-18
Use `u16::MAX` associated constant instead of from `std::u16`

faern:
What a surprise that we still used these! I was part of pushing for adding these as associated constants many years ago :D Apparently clippy recently gained this lint (?). Anyway, easy fix!

Follow up to #65 

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/67)


Serock3 (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/67)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #73, opened 2025-01-23
Run the git commit message style CI on PRs

faern:
It was previously set to run on pushes only, and that did not trigger on third party PRs, only on PRs on branches in the original repository

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/73)

### DIFF
--- a/.github/workflows/git-commit-message-style.yml
+++ b/.github/workflows/git-commit-message-style.yml
@@ -2,6 +2,7 @@
 name: Git - Check commit message style
 on:
   push:
+  pull_request:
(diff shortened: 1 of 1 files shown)


faern:
Ah, dang. I forgot that this action, if set to run on `pull_request` also unconditionally checks the PR message here on github. Which is unreasonable to follow the same formatting standards as the git commit messages. There is no sane reason to limit a line to 72 characters in a web markdown field. The automatically inserted reviewable links will also never adhere to this standard.

PR #90, opened 2025-11-28
Fix `Protocol error` by removing `NLM_F_ACK` from batch begin/end messages

Centaurus99:
Hi there,

Thank you for maintaining this crate! It has been very helpful.

I recently upgraded to `v0.9.0` and noticed that the examples are no longer running successfully. For instance, when running:

```bash
CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER='sudo -E' cargo run --example add-rules
```

I encounter the error: `Error: Os { code: 71, kind: Uncategorized, message: "Protocol error" }`

I have reproduced this issue on both Ubuntu 20.04.6 LTS (kernel 5.4.0-216 and libnftnl 1.1.5) and Ubuntu 24.04.3 LTS (kernel 6.8.0-87 and libnftnl 1.2.6).

**Investigation:**

After some digging, I found that the kernel did not support the `NLM_F_ACK` flag on batch begin/end messages until [this commit](https://github.com/torvalds/linux/commit/bf2ac490d28c21a349e9eef81edc45320fca4a3c). Consequently, on **kernels prior to 6.10**, the kernel does not reply with an ACK for batch begin/end messages even if the `NLM_F_ACK` flag is set.

This causes a mismatch in the example code's ACK handling logic. The client expects the first ACK sequence number to be 1 (for the batch begin message), but instead receives 2 (since sequence 1 was skipped by the kernel). This sequence mismatch results in the `Protocol error` seen here:
https://github.com/mullvad/nftnl-rs/blob/dd2ec80d6ce1533569b6e37d18fb98376068ac07/nftnl/examples/add-rules.rs#L197-L207

**Proposed Fix:**

This PR removes the `NLM_F_ACK` flag from batch begin/end messages and updates the `sequence_numbers` function of `FinalizedBatch` accordingly. I chose this approach for two reasons:

1.  **Compatibility:** It ensures consistent behavior across all kernel versions by uniformly treating batch begin/end messages as not expecting an ACK, avoiding the need for version-specific handling.
2.  **Alignment with upstream:** The `libnftnl` C library examples also do not set the `NLM_F_ACK` flag for these messages, as seen here: https://git.netfilter.org/libnftnl/tree/examples/nft-rule-add.c?h=libnftnl-1.2.6#n165

I would appreciate it if you could review this change and let me know if this approach looks correct to you.

Thanks again for your time and work on this project!

- - -
This change is [Reviewable](https://reviewable.io/reviews/mullvad/nftnl-rs/90)

### DIFF
--- a/nftnl/src/batch.rs
+++ b/nftnl/src/batch.rs
@@ -1,6 +1,5 @@
 use crate::{MsgType, NlMsg};
 use core::fmt;
-use nftnl_sys::libc::NLM_F_ACK;
 use nftnl_sys::{self as sys, libc};
 use std::ffi::c_void;
 use std::ops::Range;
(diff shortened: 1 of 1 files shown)


hulthe:
Thanks for the fix :)

